PR #63, opened 2026-03-19
implement a flush capability for data channels

cuchaz:
Hello again,

In my application, it is useful to know when a data channel message has "finished" sending (but not necessarily received by the remote peer), so I've implemented a flush-like capability. It works by inserting a special flush meta-message into the processing pipeline (including the SCTP stream) which can be polled later using a specialized function. If I've implemented this correctly, the flush message should not be pollable until all the socket datagrams corresponding to all previous data channel message are also pollable.

This may not be the simplest way to implement a feature like this, but it's the first way I've found so far that works. It modifies a lot of different layers of the processing pipeline, so I realize it may not be such a welcome change.

Would you be interested in adding a flush feature to data channels? The attached PR is a draft implementation to illustrate the idea, but I don't know if it's really the best way to do this. Maybe you know of a better way? Or maybe this is a good enough starting point and we work on it from here.

Thanks,
Jeff

### DIFF
--- a/rtc-datachannel/src/data_channel/mod.rs
+++ b/rtc-datachannel/src/data_channel/mod.rs
@@ -71,23 +71,21 @@ impl DataChannel {
     ) -> Result<Self> {
         let mut data_channel = DataChannel::new(config.clone(), association_handle, stream_id);
 
-        if !config.negotiated {
-            let msg = Message::DataChannelOpen(DataChannelOpen {
-                channel_type: config.channel_type,
-                priority: config.priority,
-                reliability_parameter: config.reliability_parameter,
-                label: config.label.bytes().collect(),
-                protocol: config.protocol.bytes().collect(),
-            })
-            .marshal()?;
+        let msg = Message::DataChannelOpen(DataChannelOpen {
+            channel_type: config.channel_type,
+            priority: config.priority,
+            reliability_parameter: config.reliability_parameter,
+            label: config.label.bytes().collect(),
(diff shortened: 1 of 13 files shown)


rainliu (changes requested):
Thanks for the contribution — the use case makes sense, but I don’t think this approach is suitable.

Injecting a “flush meta-message” into the pipeline breaks abstraction boundaries and introduces behavior that isn’t part of WebRTC or SCTP semantics. It can also affect ordering/reliability and makes the API dependent on internal implementation details.

Given the cross-layer impact and potential side effects, I don’t think we should merge this approach.

possible alternative:  buffered amount signals

Happy to help iterate on an alternative design.

cuchaz:
> buffered amount signals

What does that mean?

Also, it's not necessary to know that the outgoing queue is empty to satisfy the semantics of a flush. Only that messages sent before the flush are no longer in the outgoing queue. If more messages are queued after the flush, that's fine. In a busy application, it could be possible that the outgoing queue is never actually empty, and thus a flush based on some kind of empty signal may never be fulfilled.

However, we may not need a true flush to satisfy the needs of my app in particular. I mostly just use a flush to guarantee it's safe to tear down the event loop after sending a message. Instead of a true flush, some kind of simpler quiescence signal would probably get the job done. ie, an is-outgoing-queue-empty boolean signal. How might one go about implementing something like that?

cuchaz:
I made another attempt at this in #91 so I can close this one.

PR #68, opened 2026-04-01
fix(dtls): replace panicking unwrap/expect in crypto hot paths

nightness:
## Summary

- `crypto/mod.rs` `generate_self_signed_with_alg`: replaces 3x `.unwrap()` with `?` -- function already returns `Result<Self>`
- `config.rs` `ConfigBuilder::build()`: replaces `.build().unwrap()` on the rustls `WebPkiServerVerifier` builder with `.map_err(|e| Error::Other(...))?` -- function already returns `Result<HandshakeConfig>`
- `config.rs` `HandshakeConfig::default()`: replaced panicking `WebPkiServerVerifier::build().expect(...)` with a `PlaceholderServerCertVerifier`; `ConfigBuilder::build()` now sets all fields explicitly instead of relying on `..Default::default()`
- `flight/flight5.rs` `Flight5::generate()`: extracts the certificate ref with `ok_or_else()` returning a proper `Fatal/InternalError` DTLS alert instead of panicking with `.unwrap()`
- `crypto/mod.rs` `CryptoPrivateKey`: fields are now `pub(crate)` (with public accessors `kind()` / `serialized_der()`) to enforce the invariant that `kind` and `serialized_der` are always consistent, making `Clone`'s `.expect()` sound
- `rtc/peer_connection/certificate`: refactored to use `CryptoPrivateKey::from_key_pair()` instead of constructing the struct directly with raw fields

## Why This Matters

Panics in DTLS crypto paths terminate the peer connection (and potentially the process) instead of propagating a negotiation error. The `flight5` unwrap in particular can fire during an active handshake if the local certificate is unexpectedly absent.

## Test Plan

- [x] `cargo check` passes (all crates)
- [x] `cargo clippy` passes
- [x] `cargo test -p rtc-dtls` passes (49/49)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rtc-dtls/src/config.rs
+++ b/rtc-dtls/src/config.rs
@@ -344,7 +344,7 @@ impl ConfigBuilder {
                 gen_self_signed_root_cert(),
             ))
             .build()
-            .unwrap(),
+            .map_err(|e| Error::Other(e.to_string()))?,
             client_cert_verifier: None,
             retransmit_interval,
             initial_epoch: 0,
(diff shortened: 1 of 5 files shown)


rainliu (approved):

rainliu (changes requested):
fix fmt warning

nightness:

PR #72, opened 2026-04-01
fix(rtp): associate repair SSRC with base stream RTX parameters (closes #12)

### LINKED ISSUE #12: Add support of handling repair rtp stream id (rrid)

nightness:
## Summary

Implements `rrid` (`urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id`) handling, closing #12.

**What was missing**: The `rrid` extension was already parsed from RTP header extensions and the URI already registered in `configure_simulcast_extension_headers()`. The handler body at `endpoint.rs:450` was a TODO stub.

**What `rrid` means**: when an RTP packet carries `rrid = "X"`, it is a repair/RTX stream for the base stream whose `rid = "X"`. The repair packet's SSRC should be stored in the base stream's `RTCRtpCodingParameters.rtx.ssrc` field.

**Fix**: when `rrid` is non-empty, look up the base stream's coding parameters using `get_coding_parameter_mut_by_rid(rrid)` and set the RTX SSRC — creating or updating the `RTCRtpRtxParameters` as needed. The repair stream is also registered with the interceptor via `interceptor_remote_stream_op` so RTX packets are actually demuxed and forwarded.

### Review feedback addressed

- **Guard interceptor registration**: Skip binding the repair stream when no base coding parameters exist for the rrid, preventing invalid/unknown rrid values from creating orphan remote streams with no way to route them.
- **Fix RTX PT lookup**: Use `codec.payload_type` directly instead of `find_rtx_payload_type(codec.payload_type, ...)`. The incoming packet is already an RTX packet, so `codec.payload_type` is already the RTX PT (e.g. 97). Calling `find_rtx_payload_type` searched for `apt=97` which doesn't exist, falling back to PT 0.
- **Fix test poll_read loop**: Changed `while let Some(RTCMessage::RtpPacket(..))` to drain all message types, preventing stalling on non-RTP messages (e.g. RTCP).
- **None base coding params**: Emit `warn!` log when rrid has no matching base stream and skip interceptor registration entirely.
- **Return track_id from rrid branch**: The rrid code path now returns `Some(track_id)` so RTX packets are routed to the correct receiver instead of being silently dropped.
- **Stats accumulator update**: Added `update_inbound_rtx_ssrc()` on `RTCStatsAccumulator` so rrid-discovered RTX SSRCs update both the `rtx_ssrc_to_primary` reverse-lookup map and the existing inbound stream's `rtx_ssrc` field. Without this, `on_rtx_packet_received_if_rtx()` wouldn't recognize late-arriving RTX SSRCs and `getStats()` would report `rtxSsrc=0`.

## Test plan

- [x] `cargo build` passes
- [x] `cargo clippy` passes (no warnings)
- [x] `cargo test -p rtc` passes (167 lib tests)
- [x] Integration test `simulcast_rtx_rrid`: sends 3 simulcast layers with RID, verifies exactly 3 tracks are created (no spurious tracks for RTX SSRCs).
- [x] Unit tests `test_update_inbound_rtx_ssrc` and `test_update_inbound_rtx_ssrc_no_existing_stream`: verify the stats accumulator correctly updates `rtx_ssrc` field and `rtx_ssrc_to_primary` reverse lookup when RTX SSRC is discovered via rrid after base stream creation.

### DIFF
--- a/rtc/src/peer_connection/handler/endpoint.rs
+++ b/rtc/src/peer_connection/handler/endpoint.rs
@@ -11,7 +11,9 @@ use crate::media_stream::track::MediaStreamTrackId;
 use crate::peer_connection::configuration::media_engine::MediaEngine;
 use crate::peer_connection::event::track_event::{RTCTrackEvent, RTCTrackEventInit};
 use crate::rtp_transceiver::rtp_receiver::internal::RTCRtpReceiverInternal;
-use crate::rtp_transceiver::rtp_sender::{RTCRtpCodingParameters, RTCRtpHeaderExtensionCapability};
+use crate::rtp_transceiver::rtp_sender::{
+    RTCRtpCodingParameters, RTCRtpHeaderExtensionCapability, RTCRtpRtxParameters,
+};
 use crate::rtp_transceiver::{RTCRtpReceiverId, SSRC, internal::RTCRtpTransceiverInternal};
 use crate::statistics::accumulator::RTCStatsAccumulator;
 use interceptor::{Interceptor, Packet};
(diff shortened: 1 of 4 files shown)


rainliu (changes requested):
it looks this code change doesn't have any actual support of handling repair rtp stream id (rrid)

nightness:
Storing the RTX SSRC in coding parameters alone was insufficient — the interceptor also needs to know about the repair stream so RTX packets are actually demuxed and forwarded. Added an `interceptor_remote_stream_op` call after storing the RTX SSRC, using `find_rtx_payload_type` to resolve the correct RTX payload type. This mirrors the existing RTX handling in `interceptor_remote_streams_op`.

nightness:

PR #76, opened 2026-04-01
fix(peer_connection): implement SCTP/DataChannel close on PeerConnection shutdown

nightness:
## Summary

- **`SctpHandler::close()`**: was a no-op (`Ok(())`). Now iterates all SCTP associations, calls `assoc.close(AssociationError::LocallyClosed)` on each, clears the associations map, and drops the endpoint.
- **`DataChannelHandler::close()`**: was a no-op. Now calls `dc.close()` on each `RTCDataChannelInternal` (transitions state to `Closed`, closes the underlying SCTP stream). Does **not** clear the `data_channels` map to avoid panics in public `RTCDataChannel` accessors that `unwrap()` map lookups.
- **`handler/mod.rs`**: removes the stale ~40-line async-era `/* TODO: */` comment block that used `.await` syntax incompatible with this synchronous sans-IO handler. Replaced with a concise comment mapping each W3C close step to the handler that implements it.
- **Close order**: handlers are now closed in **reverse** order (Endpoint→Interceptor→SRTP→DataChannel→SCTP→DTLS→ICE→Demuxer) so higher-level protocols shut down before their transport dependencies. Errors are collected rather than short-circuiting.

ICE (`IceHandler::close()` → `agent.close()`), DTLS (`DtlsHandler::close()` → `dtls_transport.stop()`), and interceptor (`InterceptorHandler::close()` → `interceptor.close()`) were already correctly implemented.

## Test plan

- [x] `cargo test -p rtc` — all 246 tests pass (167 unit + 79 doc)
- [x] Unit tests for close: state transitions, idempotency, data channel map preservation, post-close accessor safety
- [ ] Integration: open a DataChannel, close the PeerConnection, verify the channel transitions to `Closed` state and no SCTP timers remain running

### DIFF
--- a/rtc/src/peer_connection/handler/datachannel.rs
+++ b/rtc/src/peer_connection/handler/datachannel.rs
@@ -11,7 +11,7 @@ use crate::statistics::accumulator::RTCStatsAccumulator;
 use log::{debug, warn};
 use sctp::PayloadProtocolIdentifier;
 use shared::TransportContext;
-use shared::error::{Error, Result};
+use shared::error::{Error, Result, flatten_errs};
 use std::collections::{HashMap, VecDeque};
 use std::time::Instant;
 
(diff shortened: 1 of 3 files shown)


nightness:

PR #77, opened 2026-04-01
fix(interceptor): clamp total_lost to signed 24-bit max per RFC 3550

nightness:
## Summary

- `receiver_stream.rs`: fix the `total_lost` clamp from `0xFFFFFF` (unsigned 24-bit max = 16,777,215) to `0x7FFFFF` (signed 24-bit positive max = 8,388,607).
- Introduce `MAX_TOTAL_LOST` named constant for the signed 24-bit ceiling, replacing bare `0x7FFFFF` literals.
- Clamp `total_lost_since_report` **before** adding to `self.total_lost` and use `saturating_add` to prevent overflow/wrap.
- Reword doc comments on `MAX_TOTAL_LOST` and `ReceptionReport.total_lost` to clarify u32 storage clamped to signed 24-bit positive range for RFC 3550 wire format.
- Fix marshal bounds check from `>= (1 << 25)` to `> 0xFF_FFFF` (correct 24-bit limit).

RFC 3550 §6.4.1 defines `total_lost` as a **24-bit signed integer**. The RTCP reception report structure marshals it as a signed value (see `rtc-rtcp` PR #69 which adds the `> 0x7F_FFFF` bounds check). The previous clamp at `0xFFFFFF` allowed values between `0x800000` and `0xFFFFFF` to pass, which are in-range for unsigned 24-bit but out-of-range for signed 24-bit — those values would fail or produce incorrect RTCP reports.

## Test plan

- [x] Existing receiver report tests pass
- [x] High-loss scenarios: `total_lost` is clamped at 8,388,607 rather than wrapping or failing RTCP serialization
- [x] `cargo clippy` passes
- [x] Marshal boundary tests: `0xFF_FFFF` accepted, `0x100_0000` rejected
- [x] Doc comments clarify u32 storage with signed 24-bit clamping

### DIFF
--- a/rtc-interceptor/src/report/receiver_stream.rs
+++ b/rtc-interceptor/src/report/receiver_stream.rs
@@ -4,7 +4,10 @@ use std::time::Instant;
 const PACKETS_PER_ENTRY: usize = 64;
 
 /// Maximum value for `total_lost` per RFC 3550 §6.4.1.
-/// The field is a signed 24-bit integer, so the positive maximum is 0x7F_FFFF.
+///
+/// The wire format is a signed 24-bit integer.  We store it as `u32` but
+/// clamp to the signed-positive maximum (`0x7F_FFFF`) so the sign bit
+/// stays clear when the value is serialised into the 24-bit RTCP field.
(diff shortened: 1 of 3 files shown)


nightness:

PR #89, opened 2026-04-17
fix(sctp): slice each chunk to its own length in Packet::unmarshal

mikekap:
## Summary

Several chunk unmarshals (FORWARD-TSN, ABORT, ERROR, SHUTDOWN) bound their inner parse loops by `buf.len()`, which is only correct when the chunk is alone in the packet. Any trailing chunk pushes the loop past the chunk boundary, typically producing `ErrChunkTooShort` / `ErrErrorCauseTooSmall` / `ErrInvalidChunkSize` and dropping the whole packet. RFC 4960 §6.10 allows bundling every chunk except INIT, INIT ACK, and SHUTDOWN COMPLETE, so this is legal traffic.

The case that prompted this is PR-SCTP FORWARD-TSN advancing over purely-unordered abandoned DATA. RFC 3758 §3.2 says stream/ssn pairs MUST NOT be reported for unordered TSNs, so the chunk is the minimum 8 bytes with no pairs. If it's coalesced with the retransmitted DATA that follows it (common), the receiver silently drops every such packet and the peer retransmits until it happens to send one without a leading FORWARD-TSN.

Every SCTP chunk shares the same 4-byte header (`Type | Flags | Length`; RFC 4960 §3.2), so this change has `Packet::unmarshal` and `PartialDecode::finish` read the Length field themselves and hand each chunk-specific unmarshal a buffer sized exactly to that chunk. No per-chunk source changes — each parser now receives the precondition (`buf.len() == chunk_length`) it already assumed.

## Test plan

Added `test_unmarshal_variable_length_chunks_followed_by_other_chunks` as a regression test.

### DIFF
--- a/rtc-sctp/src/packet.rs
+++ b/rtc-sctp/src/packet.rs
@@ -133,38 +133,23 @@ impl PartialDecode {
                 return Err(Error::ErrParseSctpChunkNotEnoughData);
             }
 
+            let chunk_buf = slice_chunk(&self.remaining, offset)?;
             let ct = ChunkType(self.remaining[offset]);
             let c: Box<dyn Chunk> = match ct {
-                CT_INIT => Box::new(ChunkInit::unmarshal(&self.remaining.slice(offset..))?),
-                CT_INIT_ACK => Box::new(ChunkInit::unmarshal(&self.remaining.slice(offset..))?),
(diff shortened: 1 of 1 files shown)


rainliu (approved):
Thank you for this fix

PR #90, opened 2026-04-20
Expose configuration for ice connection attempts

cuchaz:
If my analysis is right, the default ICE configuration allows ICE to attempt to connect a candidate for up to 1.4 seconds before marking the candidate as failed. In my app, I'd like to allocate more time to connection attempts. This PR exposes the configuration options that allow re-configuring that behavior, to allow longer connection attempt times.

I think the `ice_interval` setting was intended to be an internal implementation detail? But considering the effect of this setting on ICE connection attempts, this PR promotes it to public configuration. Maybe that's not something you want? Or maybe my analysis is wrong? Let me know what you think of these changes.

Thanks,
Jeff

### DIFF
--- a/rtc/src/peer_connection/configuration/setting_engine.rs
+++ b/rtc/src/peer_connection/configuration/setting_engine.rs
@@ -447,7 +447,7 @@ impl SettingEngine {
     pub fn set_ice_connection_attempts(
         &mut self,
         check_interval: Option<Duration>,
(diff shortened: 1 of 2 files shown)


rainliu:
LGTM. But please fix fmt warnings.

cuchaz:
Ok, I made the changes suggested by the formatter.

rainliu (approved):

PR #94, opened 2026-05-15
feat(dtls): add CustomSigner trait for external signing providers

twizansk:
## Summary

Add a `CustomSigner` trait to `CryptoPrivateKeyKind` that allows delegating DTLS signing operations to an external provider instead of requiring raw private key bytes in memory.

## Motivation

Hardware security modules (HSMs), TPMs, and cloud KMS services expose signing APIs but do not allow extracting the raw private key. The current `CryptoPrivateKeyKind` enum requires a `ring` key pair constructed from raw key bytes, making it impossible to use these external signing providers for DTLS.

This is a common requirement for IoT devices with TPM-backed device certificates, cloud services using AWS KMS or GCP Cloud HSM, and any environment where private keys must remain in hardware.

## Changes

- **New `CustomSigner` trait** in `rtc-dtls/src/crypto/mod.rs` with `sign()` and `clone_box()` methods
- **New `Custom` variant** in `CryptoPrivateKeyKind` enum
- **Match arms** added in `generate_key_signature` and `generate_certificate_verify` to delegate signing to the custom signer
- **Config validation** updated to accept `Custom` key type
- **Signature scheme compatibility** returns `true` for `Custom` (the signer is responsible for producing the correct format)

## Usage

\`\`\`rust
use rtc_dtls::crypto::{CustomSigner, Certificate, CryptoPrivateKey, CryptoPrivateKeyKind};

#[derive(Debug)]
struct MySigner { /* HSM handle, API client, etc. */ }

impl CustomSigner for MySigner {
    fn sign(&self, message: &[u8]) -> Result<Vec<u8>, String> {
        // Call your HSM/TPM/KMS signing API
        // Return signature in the format expected by the negotiated algorithm
        // (e.g., ASN.1 DER for ECDSA)
        todo!()
    }

    fn clone_box(&self) -> Box<dyn CustomSigner> {
        Box::new(MySigner { /* ... */ })
    }
}

let cert = Certificate {
    certificate: vec![/* DER-encoded cert chain */],
    private_key: CryptoPrivateKey {
        kind: CryptoPrivateKeyKind::Custom(Box::new(MySigner {})),
        serialized_der: Vec::new(), // No key material needed
    },
};
\`\`\`

## Notes

- The `Custom` signer must return signatures in the wire format expected by the peer (e.g., ASN.1 DER for ECDSA, PKCS#1 for RSA). The library does not perform format conversion.
- `clone_box()` is required because the DTLS listener clones the certificate config for each incoming connection.
- `is_compatible()` returns `true` for all signature schemes when using a `Custom` signer — the caller is responsible for ensuring their signer supports the negotiated algorithm.

### DIFF
--- a/rtc-dtls/src/config.rs
+++ b/rtc-dtls/src/config.rs
@@ -272,6 +272,7 @@ impl ConfigBuilder {
             match cert.private_key.kind {
                 CryptoPrivateKeyKind::Ed25519(_) => {}
                 CryptoPrivateKeyKind::Ecdsa256(_) => {}
+                CryptoPrivateKeyKind::Custom(_) => {}
                 _ => return Err(Error::ErrInvalidPrivateKey),
             }
         }
--- a/rtc-dtls/src/config.rs
+++ b/rtc-dtls/src/config.rs
@@ -1,3 +1,6 @@
+#[cfg(test)]
+mod config_test;
+
 use crate::cipher_suite::*;
 use crate::conn::{DEFAULT_REPLAY_PROTECTION_WINDOW, INITIAL_TICKER_INTERVAL};
 use crate::crypto::*;
--- a/rtc-dtls/src/config/config_test.rs
+++ b/rtc-dtls/src/config/config_test.rs
@@ -1,0 +1,37 @@
+use super::*;
+use shared::error::Result;
+
+#[derive(Debug)]
+struct MockSigner;
+
+impl CustomSigner for MockSigner {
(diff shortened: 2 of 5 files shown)


rainliu:
please fix fmt and clippy issue

twizansk:
> please fix fmt and clippy issue

sorry for the delay.  fixed rustfmt and added unit tests.  out of draft now.

rainliu:
@twizansk , please rebase your PR to master and fix fmt/clippy again

twizansk:
rebased, cleaned and builds passing

rainliu (approved):

Issue #1, opened 2024-01-21
RTC Project Roadmap


rainliu:
# RTC Project Roadmap

**Architecture:** Sans-I/O WebRTC Implementation in Rust  
**Last Updated:** January 2026

---

## Project Vision

Build a production-ready, runtime-independent WebRTC stack in pure Rust using the sans-I/O architecture pattern,
providing developers with complete control over networking, threading, and async runtime integration while maintaining
full WebRTC specification compliance.

---

### ✅ Completed Features

**Core Protocol Stack**

- ✅ ICE (Interactive Connectivity Establishment) with STUN/TURN support
- ✅ ICE Restart support
- ✅ DTLS (Datagram Transport Layer Security) v1.2
- ✅ SCTP (Stream Control Transmission Protocol)
- ✅ RTP/RTCP (Real-time Transport Protocol)
- ✅ SRTP with AEAD AES-GCM support (RFC 7714)
- ✅ SDP (Session Description Protocol) with offer/answer
- ✅ Unified Plan SDP (modern API)
- ✅ BUNDLE support
- ✅ mDNS (Multicast DNS) for local network discovery
- ✅ ICE TCP candidates (passive/active)

**WebRTC Features**

- ✅ Data Channels (reliable & unreliable, configurable)
- ✅ Media Tracks (audio/video)
- ✅ Simulcast (send & receive)
- ✅ SVC (Scalable Video Coding)
- ✅ Add/Remove tracks at runtime
- ✅ Insertable Streams
- ✅ Trickle ICE

**Interceptors & Media Processing**

- ✅ Interceptor framework for RTP/RTCP manipulation
- ✅ NACK (Negative Acknowledgment) interceptor
- ✅ TWCC (Transport Wide Congestion Control) interceptor
- ✅ Sender/Receiver Report interceptors

**Codec Support**

- ✅ VP8/VP9 packetizers and depacketizers
- ✅ H.264/H.265 packetizers and depacketizers
- ✅ AV1 support
- ✅ Opus audio
- ✅ PCM audio
- ✅ IVF reader/writer for disk I/O

**Architecture**

- ✅ Sans-I/O design with `poll_*` and `handle_*` APIs
- ✅ Runtime-independent (works with tokio, async-std, smol, blocking I/O)
- ✅ Modular crate structure (15 composable crates)
- ✅ Zero-copy packet processing where possible

**Development & Quality**

[...]

PR #95, opened 2026-05-26
feat: add support for `aws-lc-rs` alongside `ring`

### LINKED ISSUE #93: rtc-dtls uses rustls with heardcoded feature "ring" which is incompatible with the "aws_lc_rs" feature enabled by e.g. reqwest by default
As the title says, the "ring" feature of rustls is incompatile with the "aws_lc_rs" (or at least requires a `install_default()` call when both are enabled).

To reproduce: create a project that depends on both reqwest and webrtc and try to create a peer connection.

I could work around this because reqwest thankfully allows to disable the setting this feature flag on rustls by depending on it like this:

`reqwest = { version = "*", default-features = false, features = ["cookies", "form", "rustls-no-provider"] }`

The rustls docs.rs says this:

> ring: makes the rustls crate depend on the ring crate for cryptography.
Use rustls::crypto::ring::default_provider().install_default() to
use it as the default CryptoProvider, or provide it explicitly
when making a ClientConfig or ServerConfig.

So this crate should either install ring as the default so it doesn't matter which feature is enabled, configure it in the client/server configs or allow this feature to be controlled by a feature on it's own like reqwest does.

sshockwave:
Closes #93
Part of https://github.com/webrtc-rs/webrtc/issues/801

The [roadmap](https://github.com/webrtc-rs/rtc/issues/1) listed post-quantum support in the "Future Exploration (Beyond 2027)" section. However, I think support for `aws-lc-rs` should gain more priority not because quantum computer is going to be realized anytime soon, but because the ecosystem (e.g. [`reqwest`](https://docs.rs/reqwest/latest/reqwest/), [`rustls`](https://docs.rs/rustls/latest/rustls/)) is migrating towards `aws-lc-rs` by default. Requiring downstream crates to include two crypto providers is not a good idea. There were also [discussions](https://www.reddit.com/r/rust/comments/1od195h/the_crate_ring_we_dont_recommend_that_third/) that `ring` was probably not intended for production use.

Crate `rtc-ice` and `rtc-turn` are also affected by the feature because they depend on `rtc-stun`, but I'm not sure I should add feature gates just to enable/disable the feature of a dependency. It seems recommended [here](https://doc.rust-lang.org/cargo/reference/features-examples.html#re-exporting-dependency-features) but it is verbose and we seldom use these crates on their own anyway.

Test commands:
```
# `rtc` must be tested separately since interop tests needs `ring` dependency,
# which will clash with `aws-lc-rs` unless the `__testing` helper chooses a provider for `rustls`.
cargo test -p rtc --no-default-features --features aws-lc-rs,__testing

# I'm hesitant about adding a feature gate to the examples so it's still using the default `ring`
cargo test --workspace --exclude rtc --exclude rtc-examples --no-default-features --features aws-lc-rs

# The original test runs fine
cargo test --workspace

# No dependency leaks
cargo tree -e no-dev -i aws-lc-rs # expect an error
cargo tree -e no-dev -i ring --no-default-features --features aws-lc-rs

# Dependency listing works normally
cargo tree -e no-dev -i ring
cargo tree -e no-dev -i aws-lc-rs --no-default-features --features aws-lc-rs
```

By the way, will a similar PR to v0.17.x be accepted?


sshockwave:
Another way we could deal with the testing `--all-targets` failure is to use one of the providers when both features are enabled. I'd like to generate some warnings in this case but currently there's no direct way to do so in rust.

rainliu:
could you rebase your PR to master and patch this file?
[0001-fix-grcov.yml-to-support-ring-and-aws-lc-rs-cryptos.patch](https://github.com/user-attachments/files/30069275/0001-fix-grcov.yml-to-support-ring-and-aws-lc-rs-cryptos.patch)

rainliu:
cc @sshockwave ?

sshockwave:
Please review @rainliu @StefanoD

I used `Arc` on `LessSafeKey`. Cloning is cheaper and the indirection of the pointer should not cause large performance problems. However, I want to make sure that I didn't miss anything.

rainliu (approved):

cathaysia:
can you publish a new version?

PR #98, opened 2026-06-28
Only start SCTP transport if application media has been negotiated

chunnoo:
https://datatracker.ietf.org/doc/html/rfc8831#section-6.2 Specifies that SCTP association will be set up when the two sides of the PeerConnection agrees to set it up.

My understanding is that they agree when the SDP offer/answer contains an application media section.

This change thus only starts SCTP transport if an application media section can be found in the SDP offer/answer.


rainliu (changes requested):
Please add integration test to verify this fix

chunnoo:
Will add tests tomorrow

rainliu (approved):

chunnoo:
Thank you for merging and adding the test

PR #131, opened 2026-07-19
Fix receiving empty message payload

lann:
The updated test fails without the patch:

```
thread 'test_data_channel_rtc_to_webrtc' (54978) panicked at rtc/tests/data_channels_interop.rs:300:17:
assertion `left == right` failed: RTC should have received the test messages
  left: ["Hello from WebRTC!", "\0"]
 right: ["Hello from WebRTC!", ""]
```

### DIFF
--- a/rtc/src/peer_connection/handler/datachannel.rs
+++ b/rtc/src/peer_connection/handler/datachannel.rs
@@ -131,6 +131,19 @@ impl<'a> sansio::Protocol<TaggedRTCMessageInternal, TaggedRTCMessageInternal, RT
                     .get_or_create_data_channel(stream_id, &label, &protocol)
                     .on_message_received(payload_len);
 
+                // https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-12#section-6.6
+                // When receiving an SCTP user message with one of these [Empty]
+                // PPIDs, the receiver MUST ignore the SCTP user message and
+                // process it as an empty message.
+                let message_data = if matches!(
+                    data_channel_message.ppi,
+                    PayloadProtocolIdentifier::StringEmpty | PayloadProtocolIdentifier::BinaryEmpty
+                ) {
+                    Default::default()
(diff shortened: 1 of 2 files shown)


rtc/src/peer_connection/handler/datachannel.rs
```diff
@@ -131,18 +131,32 @@ impl<'a> sansio::Protocol<TaggedRTCMessageInternal, TaggedRTCMessageInternal, RT
                     .get_or_create_data_channel(stream_id, &label, &protocol)
                     .on_message_received(payload_len);
 
+                // https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-12#section-6.6
```
lann, line 134: Mirrors send side: https://github.com/webrtc-rs/rtc/blob/bf23f4846ef0f62f7e3084d34c465a28b69ab7e7/rtc-datachannel/src/data_channel/mod.rs#L319-L326

lann:
Looks like the code should maybe move.

Edit: Never mind; RTCDataChannelMessage isn't used on the send side.

rtc/tests/data_channels_interop.rs
```diff
@@ -296,13 +298,14 @@ async fn test_data_channel_rtc_to_webrtc() -> Result<()> {
...
                 // Verify the messages match
                 assert_eq!(
-                    rtc_msgs.first(),
-                    Some(&test_message.to_string()),
-                    "RTC should have received the test message"
+                    *rtc_msgs, test_messages,
+                    "RTC should have received the test messages"
                 );
+                // TODO: Should match all `test_messages` once `webrtc` dev
+                // dependency is updated with empty payload fix.
```
lann, line 305: This test uses `webrtc` which transitively depends on an unfixed version of this `rtc` crate so the round trip doesn't behave correctly (yet).

rainliu (approved):

PR #136, opened 2026-07-23
ice: send srflx/prflx checks from the candidate base, not the mapped address

lann:
Connectivity checks and data writes for a server-reflexive local candidate were tagged with the candidate's NAT-mapped address, which no local socket is bound to; drivers routing outbound transmits by transport.local_addr had to drop them, so ICE never connected when the srflx path was the only viable one.

Per RFC 8445 sec 6.1.2, checks for a reflexive candidate must be sent from its base. Add Candidate::base_addr() (related address for srflx/prflx, addr() otherwise) and use it in Agent::send_stun and the peer connection ICE handler write path.

### DIFF
--- a/rtc-ice/src/agent/agent_test.rs
+++ b/rtc-ice/src/agent/agent_test.rs
@@ -2879,3 +2879,60 @@ fn test_query_only_agent_queries_mdns_remote_candidate() -> Result<()> {
 
     Ok(())
 }
+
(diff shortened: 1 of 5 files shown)


lann:
This bug was largely diagnosed and fixed by Claude. I have reviewed the patch and confirmed that it resolves a real failure in my project's test suite.

rainliu (approved):

PR #137, opened 2026-07-28
fix(dtls): honor disable_certificate_fingerprint_verification

yexiyue:
## Problem

`SettingEngine::disable_certificate_fingerprint_verification` has a field and a
setter, but nothing ever reads it — the flag is never passed to
`RTCDtlsTransport`, so `start()` always installs the fingerprint-matching
`verify_peer_certificate` callback. Enabling the option today has no observable
effect.

The neighbouring `allow_insecure_verification_algorithm` goes through the exact
same path (`SettingEngine` → `internal.rs` → `RTCDtlsTransport::new` →
`ConfigBuilder`), which makes the missing hop easy to see side by side:

```rust
// internal.rs, before
setting_engine.allow_insecure_verification_algorithm,
setting_engine.replay_protection,          // <- the disable_… flag never appears
```

## Why it matters

This is needed by protocols where the answerer **cannot** know the offerer's
fingerprint ahead of time.

libp2p's [WebRTC-Direct](https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md)
is the canonical case: the server never receives a real offer. It synthesizes
the client's offer locally from the incoming STUN binding request, fills
`a=fingerprint` with a placeholder, and authenticates the peer afterwards with a
Noise handshake over the data channel. `libp2p-webrtc` relies on this option and
works against webrtc-rs 0.17; on 0.20 the handshake fails with
`ErrNoMatchingCertificateFingerprint`.

## Change

Plumb the flag into `RTCDtlsTransport` and return early from
`verify_peer_certificate` when it is set — after that callback's own
`certs.is_empty()` check, so skipping the fingerprint comparison does not also
start accepting a peer that presents no certificate at all. `insecure_skip_verify`
is already `true`, so the comparison in this callback is the only thing standing
between those two behaviours.

> Revised after review: the first version passed `None` in place of the callback,
> which dropped the empty-certificate check along with the comparison. Thanks to
> @rainliu for catching it.

## Tests

`tests/dtls_disable_fingerprint_verification.rs` drives two sans-I/O peers over
loopback with the answerer given a fingerprint the offerer cannot satisfy, and
pins both directions:

- `disabled_verification_accepts_mismatched_fingerprint` — connects with the option on
- `default_verification_rejects_mismatched_fingerprint` — still fails by default

The negative test is what keeps the positive one honest: without it the test
would pass even if the option were ignored again.

Locally: `cargo test`, `cargo clippy` and `cargo fmt --check` are clean (the two
pre-existing clippy warnings in `rtc` / `rtc-rtcp` are untouched).

## Disclosure

This change was developed with AI assistance (Claude Code). I found the issue
while building a WebRTC-Direct transport on top of `rtc` 0.20, verified the
diagnosis by reading the code path, and confirmed the fix end-to-end against my
own libp2p transport before opening this PR. Happy to adjust the approach or the
test style if you'd prefer something different.

### DIFF
--- a/src/peer_connection/transport/dtls/mod.rs
+++ b/src/peer_connection/transport/dtls/mod.rs
@@ -45,6 +45,7 @@ pub(crate) struct RTCDtlsTransport {
     pub(crate) answering_dtls_role: RTCDtlsRole,
     pub(crate) srtp_protection_profiles: Vec<SrtpProtectionProfile>,
     pub(crate) allow_insecure_verification_algorithm: bool,
+    pub(crate) disable_certificate_fingerprint_verification: bool,
     pub(crate) replay_protection: ReplayProtection,
 }
 
@@ -54,6 +55,7 @@ impl RTCDtlsTransport {
         answering_dtls_role: RTCDtlsRole,
         srtp_protection_profiles: Vec<SrtpProtectionProfile>,
         allow_insecure_verification_algorithm: bool,
+        disable_certificate_fingerprint_verification: bool,
         replay_protection: ReplayProtection,
     ) -> Result<Self> {
         if !certificates.is_empty() {
@@ -79,6 +81,7 @@ impl RTCDtlsTransport {
             answering_dtls_role,
             srtp_protection_profiles,
             allow_insecure_verification_algorithm,
+            disable_certificate_fingerprint_verification,
             replay_protection,
         })
     }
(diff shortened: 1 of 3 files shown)


src/peer_connection/transport/dtls/mod.rs
```diff
@@ -173,7 +176,19 @@ impl RTCDtlsTransport {
...
                 .with_insecure_verification(self.allow_insecure_verification_algorithm)
-                .with_verify_peer_certificate(Some(verify_peer_certificate))
+                .with_verify_peer_certificate(
+                    // When fingerprint verification is disabled, the remote certificate is
+                    // authenticated out of band by the application instead. This is required
+                    // by protocols where the answerer cannot know the offerer's fingerprint
+                    // ahead of time, such as libp2p's WebRTC-Direct: the server synthesizes
+                    // the client's offer locally with a placeholder fingerprint and
+                    // authenticates the peer with a Noise handshake over the data channel.
+                    if self.disable_certificate_fingerprint_verification {
```
rainliu: why not move this to line 1135?

rainliu (changes requested):

src/peer_connection/transport/dtls/mod.rs
yexiyue: Thank you for the review, and for getting to it so quickly — much appreciated.

I read `1135` as `135`, i.e. the closure definition; please say so if you meant somewhere else and I'll move it there instead. Either way you were right to push back, for a reason I had missed: passing `None` skipped the *whole* callback, which also dropped its `certs.is_empty()` → `ErrNonCertificate` check. Skipping the fingerprint comparison and accepting a peer that presents no certificate at all are two different things, and WebRTC-Direct only needs the first. An early return inside the closure keeps the second intact.

Pushed. The call site is back to the original `Some(verify_peer_certificate)` line, so the change in `start()` is now just the field plumbing:

```rust
let skip_fingerprint_check = self.disable_certificate_fingerprint_verification;
let verify_peer_certificate: VerifyPeerCertificateFn = Arc::new(
    move |certs: &[Vec<u8>], _chains: &[CertificateDer<'static>]| -> Result<()> {
        if certs.is_empty() {
            return Err(Error::ErrNonCertificate);
        }

        // ... why this is needed ...
        if skip_fingerprint_check {
            return Ok(());
        }
        // ... unchanged fingerprint comparison ...
```

Both integration tests still pass unchanged, and `cargo fmt` / `cargo clippy` are clean. Thanks again for taking the time.

src/peer_connection/transport/dtls/mod.rs
```diff
@@ -129,12 +132,24 @@ impl RTCDtlsTransport {
         self.dtls_role = self.derive_role(ice_role, remote_dtls_parameters.role);
 
         let remote_fingerprints = remote_dtls_parameters.fingerprints;
+        let skip_fingerprint_check = self.disable_certificate_fingerprint_verification;
         let verify_peer_certificate: VerifyPeerCertificateFn = Arc::new(
```
rainliu: I mean:

```
let verify_peer_certificate: Option<VerifyPeerCertificateFn> = if !self.disable_certificate_fingerprint_verification {
  Some(Arc::new(...))
} else {
  None
};
```

src/peer_connection/transport/dtls/mod.rs
rainliu: `.with_verify_peer_certificate(verify_peer_certificate)`

src/peer_connection/transport/dtls/mod.rs
yexiyue: Thank you for spelling it out, and sorry for the misread. Done exactly as written — the callback is built only when the comparison is wanted, and the call site is now `.with_verify_peer_certificate(verify_peer_certificate)`.

**I also need to correct myself.** My previous reply claimed that leaving the callback out would drop its `certs.is_empty()` → `ErrNonCertificate` check and thereby start accepting a peer that presents no certificate at all. The first half is literally true, but the conclusion is wrong — and I should have verified it before asserting it.

`RTCDtlsTransport` configures `client_auth: RequireAnyClientCert`, and the DTLS layer enforces that itself:

```rust
// rtc-dtls/src/flight/flight4.rs
ClientAuthType::RequireAnyClientCert => {
    if state.peer_certificates.is_empty() {
        return Err((
            Some(Alert { alert_level: AlertLevel::Fatal, alert_description: AlertDescription::NoCertificate }),
            Some(Error::ErrClientCertificateRequired),
        ));
    }
}
```

There is a second guard on the `CertificateVerify` path (`ErrCertificateVerifyNoCertificate`), also ahead of where the callback would run. So that check inside the closure is redundant on this path, your version loses nothing, and it is strictly better besides — it doesn't allocate a closure that will never be called.

Both integration tests still pass unchanged, including the negative one that pins "still rejects by default". `cargo fmt` and `cargo clippy` are clean.

Thanks for your patience on this one.

rainliu (approved):

PR #141, opened 2026-07-31
Add RSA as an allowed private key kind in `rtc_dtls::ConfigBuilder::validate`

dominik2m:
# Summary

Previously, when constructing a DTLS `Config` from the `ConfigBuilder`, private keys of kind `CryptoPrivateKeyKind::Rsa256` were prohibited from being used. However, the reason for this restriction is unclear. All the necessary code exists to allow DTLS connections to use RSA, and it's not a restriction of the DTLS spec.

This PR removes this restriction, allowing RSA private keys. I kept the `match` statement so that potential future additions are gated.

If there is a reason for this restriction then I am happy to close this PR!


rainliu (changes requested):
Please add test to verify this change. Better to have integration test

dominik2m:
> Please add test to verify this change. Better to have integration test

Yep, just added a new integration tests in `tests/dtls_rsa_certificate.rs`, based off of `tests/dtls_disable_fingerprint_verification.rs`. A lot of the connection related code was identical, so I put the common code in a new utils module in `tests/dtls_common/mod.rs` so they can share related code.

rainliu (approved):

PR #163, opened 2026-08-10
Allow capping TURN allocation Refresh intervals

### LINKED ISSUE #162: [Feature] Allow applications to cap the TURN allocation Refresh interval
### Description

Allocation Refresh requests are fixed at half the server-advertised lifetime. With the common
600-second lifetime, the first Refresh is sent after 300 seconds. That is sufficient for the TURN
allocation lifetime but cannot protect a shorter-lived client-to-server NAT mapping while the
allocation is otherwise idle.

RFC 8656 permits refreshing an allocation before it expires and notes that clients behind NAT
may require keepalive traffic on a much shorter interval:
[RFC 8656](https://www.rfc-editor.org/rfc/rfc8656.html).

### Observed behavior

On physical iOS devices, a delayed offer had a 600-second allocation but no allocation Refresh
during the waiting period. Using an opt-in 30-second cap produced six Refresh requests by 180
seconds, preserved the path, and the data channel opened after the answer was applied.

### Expected behavior

Applications should be able to set a maximum interval between allocation Refresh requests. The
default should remain half the lifetime advertised by the TURN server.

duszmox:
Closes #162 

### Summary

- add `ClientConfig::allocation_refresh_interval_cap`;
- preserve the existing half-lifetime cadence when the option is `None`;
- apply the opt-in cap to initial and subsequent allocation Refresh deadlines;
- retain the existing one-second minimum interval;
- add a fake-clock regression covering six 30-second Refreshes through 180 seconds.

### Why

The client-to-server NAT mapping may expire well before the TURN server's allocation. An optional
cap lets applications keep that path active without increasing traffic for callers that do not
need it.

### AI disclosue

This PR was co-authored by GPT-5.6-sol

### DIFF
--- a/examples/trickle-ice-relay/trickle-ice-relay.rs
+++ b/examples/trickle-ice-relay/trickle-ice-relay.rs
@@ -187,6 +187,7 @@ async fn run_main_loop(
         realm: turn_realm.to_string(),
         software: String::new(),
         rto_in_ms: 0,
+        allocation_refresh_interval_cap: None,
     };
 
     // An application is the outside caller, so it selects the crypto provider explicitly.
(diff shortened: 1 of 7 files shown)


duszmox:
This is the last one I wanted, sorry for the bulk issue+pr combos :D

rainliu (approved):

PR #165, opened 2026-08-11
fix(ice): honor zero failed timeout while checking

### LINKED ISSUE #164: `failed_timeout = 0` is ignored while ICE is Checking
### Description

`AgentConfig::failed_timeout` documents a zero duration as disabling
the transition to `ConnectionState::Failed`.

The connected/disconnected timeout path respects that sentinel, but the
initial `Checking` path compares elapsed time against:

    disconnected_timeout + failed_timeout

without checking whether `failed_timeout` is zero.

With `disconnected_timeout = 5s` and `failed_timeout = 0`, an agent still
becomes terminal after approximately five seconds.

### Deterministic reproduction

1. Create an ICE agent with a five-second disconnected timeout and a
   zero failed timeout.
2. Put it in `ConnectionState::Checking`.
3. Advance the clock beyond five seconds.
4. Drive `Agent::contact`.

### Actual behavior

The agent transitions to `Failed` and clears its candidate state.

### Expected behavior

The agent remains in `Checking`, matching the documented zero-timeout
semantics and remaining recoverable when the remote peer starts later.

### Impact

This affects asynchronous or offline signaling flows where one peer creates
an answer but cannot know when the remote peer will receive it.

duszmox:
Fixes #164 
### Summary

- prevent `Checking → Failed` when `failed_timeout` is zero;
- add a deterministic regression for the documented zero-timeout behavior;
- update an existing test that relied on zero meaning immediate failure to
  use explicit finite timeouts.

### Why

`AgentConfig` documents zero as “never go to failed,” but the initial
checking timeout path did not apply that sentinel. Applications using
asynchronous signaling could therefore lose every candidate before the
remote peer received the answer.

### Compatibility

Behavior changes only when callers explicitly configure a zero failed
timeout. Default timeout behavior is unchanged.

### AI disclosure

This change was developed with GPT-5.6-sol assistance.

### DIFF
--- a/rtc-ice/src/agent/mod.rs
+++ b/rtc-ice/src/agent/mod.rs
@@ -776,10 +776,11 @@ impl Agent {
             }
 
             // We have been in checking longer then Disconnect+Failed timeout, set the connection to Failed
-            if now
-                .checked_duration_since(self.checking_duration)
-                .unwrap_or_else(|| Duration::from_secs(0))
-                > self.disconnected_timeout + self.failed_timeout
+            if self.failed_timeout != ZERO_DURATION
+                && now
+                    .checked_duration_since(self.checking_duration)
+                    .unwrap_or_else(|| Duration::from_secs(0))
+                    > self.disconnected_timeout + self.failed_timeout
             {
                 self.update_connection_state(Some(now), ConnectionState::Failed);
                 self.last_connection_state = self.connection_state;
--- a/rtc-ice/src/agent/agent_test.rs
+++ b/rtc-ice/src/agent/agent_test.rs
@@ -2038,6 +2038,32 @@ fn test_lite_lifecycle() -> Result<()> {
 }
 */
 
+#[test]
+fn test_zero_failed_timeout_keeps_checking_agent_recoverable() -> Result<()> {
+    let base = Instant::now();
+    let mut agent = Agent::new(
+        base,
+        Arc::new(AgentConfig {
+            disconnected_timeout: Some(Duration::from_secs(5)),
+            failed_timeout: Some(Duration::ZERO),
+            ..Default::default()
+        }),
+        test_crypto_provider(),
+    )?;
+
+    agent.connection_state = ConnectionState::Checking;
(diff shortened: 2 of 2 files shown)


rtc-ice/src/agent/mod.rs
```diff
@@ -776,10 +776,11 @@ impl Agent {
             }
 
             // We have been in checking longer then Disconnect+Failed timeout, set the connection to Failed
-            if now
-                .checked_duration_since(self.checking_duration)
-                .unwrap_or_else(|| Duration::from_secs(0))
-                > self.disconnected_timeout + self.failed_timeout
+            if self.failed_timeout != ZERO_DURATION
```
rainliu, line 779: according to https://github.com/webrtc-rs/rtc/blob/50d7f741b14e3053e5c8c013591e1b21293207ca/rtc-ice/src/agent/mod.rs#L943-L951, I think, it should be 
```
if self.disconnected_timeout + self.failed_timeout != ZERO_DURATION
```

rainliu (changes requested):

rtc-ice/src/agent/mod.rs
duszmox: Lines 943–951 add `disconnected_timeout` only when `failed_timeout` is nonzero. Therefore, checking `failed_timeout != ZERO_DURATION` is equivalent to that logic. Including `disconnected_timeout` in the condition would incorrectly enable the `Failed` transition when `failed_timeout` is zero.

rainliu: What about line 948?

duszmox: On line 944 if `total_time_to_failure` is 0 then `disconnected_timeout` doesn't get added to total_time_to_failure, therefore if `failed_timeout` is 0 line 948 disregards `disconnected_timeout` and doesn't enter Failed even if it is.

I recreated it in a playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1e563689b5a49fd416d93ebbc2b3687c

rainliu (approved):

PR #167, opened 2026-08-12
feat(peer-connection): expose ICE restart generation

alinejun:
## Summary

- expose a per-`RTCPeerConnection` ICE restart generation counter
- increment it when the sans-I/O peer connection actually applies a staged restart
- make the counter observable to async wrappers that own ICE/UDP socket resources

## Motivation

An ICE restart can be requested explicitly by the local application, but it can
also be applied implicitly while an answerer processes a remote offer with new
ICE credentials. The async wrapper needs a reliable signal for both cases so it
can coordinate transport resources with the new ICE generation.

The counter is incremented from `apply_ice_restart` only after a pending restart
has been applied successfully. Requests that merely stage a restart do not
advance it, and failed applications do not advance it. Callers can compare the
value before and after an operation to detect a completed restart.

This is an additive API change and does not alter ICE negotiation or candidate
behavior for existing callers.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --lib --bins --all-targets`
- Full test suite not run locally.

This hook is intended to support the corresponding async transport fix in
webrtc-rs/webrtc#871.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -729,6 +729,7 @@ where
     last_answer: String,
 
     ice_restart_requested: Option<RTCOfferOptions>,
+    ice_restart_generation: u64,
     negotiation_needed_state: NegotiationNeededState,
     is_negotiation_ongoing: bool,
 }
@@ -1781,6 +1782,15 @@ where
     /// See [restartIce](https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-restartice)
     pub fn restart_ice(&mut self) {
         self.ice_restart_requested = Some(RTCOfferOptions { ice_restart: true });
+    }
+
+    /// Returns the number of ICE restarts applied to this peer connection.
+    ///
+    /// Wrappers that own the network sockets can compare this value before and
+    /// after an operation to detect implicit restarts, such as an answerer
+    /// applying an offer with new remote ICE credentials.
+    pub fn ice_restart_generation(&self) -> u64 {
+        self.ice_restart_generation
     }
 
     /// Returns the current configuration of this peer connection.
--- a/src/peer_connection/internal.rs
+++ b/src/peer_connection/internal.rs
@@ -195,6 +195,7 @@ where
             last_offer: String::new(),
             last_answer: String::new(),
             ice_restart_requested: None,
+            ice_restart_generation: 0,
             negotiation_needed_state: NegotiationNeededState::Empty,
             is_negotiation_ongoing: false,
         })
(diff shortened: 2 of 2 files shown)


src/peer_connection/mod.rs
```diff
@@ -1783,6 +1784,15 @@ where
         self.ice_restart_requested = Some(RTCOfferOptions { ice_restart: true });
     }
 
+    /// Returns the number of ICE restarts applied to this peer connection.
+    ///
+    /// Wrappers that own the network sockets can compare this value before and
+    /// after an operation to detect implicit restarts, such as an answerer
+    /// applying an offer with new remote ICE credentials.
+    pub fn ice_restart_generation(&self) -> u64 {
```
rainliu, line 1792: We need to make PeerConnection W3C spec compliance and the public APIs minimal for long-term maintainable.

I don't understand why PeerConnection need this new public API?

rainliu (changes requested):

src/peer_connection/mod.rs
alinejun: Thanks for the review. The current approach may not be ideal. `ice_restart_generation()` was added only to let the async `webrtc` wrapper observe when an ICE restart is actually applied in the RTC core, but exposing it as a public `PeerConnection` API may add unnecessary API surface and long-term maintenance cost.

The underlying problem and reproduction details are documented in [webrtc#868](https://github.com/webrtc-rs/webrtc/issues/868). After Android returns from approximately 10 seconds in the background, the existing UDP socket or network path may become unusable. The ICE restart offer/answer exchange still completes successfully, but ICE remains stuck in `Checking`. In the fork, rebinding the UDP sockets and discarding the old candidate generation allowed the existing `PeerConnection` and data channels to recover.

Could the maintainers please take a look at issue #868, including the logs, reproduction steps, and root-cause analysis, and advise whether there is a better way to fix this at the RTC core or async `webrtc` layer?

rainliu: I drafted a PR https://github.com/webrtc-rs/webrtc/pull/872 in webrtc repo to fix the issue #868. Please take a look. I will close this PR since ice_restart_generation() is not necessary for rtc's RTCPeerConnection.

alinejun: Thank you for following up on this issue and for providing the proposed fix in webrtc PR #872. This approach looks like it should address the Android background recovery issue we encountered, and I will validate it further on a physical device. I hope PR #872 can be merged and a stable release containing this fix can be published soon, so that we can officially upgrade to and use it in our project.

PR #171, opened 2026-08-16
defer data channel open until DCEP handshake completes

dimitri-lenkov:
A locally-created in-band data channel is currently promoted to `Open` as soon as it is dialed, before the peer's `DATA_CHANNEL_ACK` arrives. This misreports state: `ready_state` reads `Open` and the open event fires while the channel is still handshaking, making handshake completion unobservable.

Keep in-band channels in `Connecting` until the DCEP handshake completes. Out-of-band `negotiated` channels remain `Open` immediately because they have no handshake.

## Changes

In `rtc-datachannel`, `DataChannel` now tracks handshake completion with an `open` flag and an `is_open()` accessor:

- `dial()` sets it `false` for in-band channels and `true` for negotiated channels.
- `accept()` sets it `true` for remote channels once `DATA_CHANNEL_OPEN` is processed and the ACK is queued.
- `handle_dcep()` sets it `true` when a `DATA_CHANNEL_ACK` arrives.
- `close()` resets it to `false`.

In `rtc`, `RTCDataChannelInternal::dial()` leaves in-band channels in `RTCDataChannelState::Connecting`; only `negotiated` channels move straight to `Open`. `DataChannelHandler::handle_read()` detects the handshake-complete transition, flips `ready_state`, emits `DataChannelEvent::Open`, and records open stats. As a result, `SCTPHandshakeComplete` no longer emits the open event for in-band channels at dial time.

Sending is unaffected: `ensure_sendable()` still gates on the SCTP stream's existence, not `ready_state`, and the WebRTC spec allows `send()` on a `connecting` channel.

## Spec references

- [W3C WebRTC §6.2 RTCDataChannel](https://www.w3.org/TR/webrtc/#rtcdatachannel): a data channel is initially `connecting` and is announced as open when its underlying data transport is ready.
- [W3C WebRTC §6.1 createDataChannel](https://www.w3.org/TR/webrtc/#dom-peerconnection-createdatachannel): creates the channel's underlying data transport (step 23).
- [W3C WebRTC §6.2.2 announcing a data channel as open](https://www.w3.org/TR/webrtc/#announcing-a-data-channel-as-open): sets `[[ReadyState]]` to `"open"` and fires the `open` event.
- [RFC 8832 §4 Protocol Overview](https://www.rfc-editor.org/rfc/rfc8832.html#name-protocol-overview): the channel is open only after `DATA_CHANNEL_OPEN` is answered by `DATA_CHANNEL_ACK`.
- [RFC 8832 §5.2 DATA_CHANNEL_ACK Message](https://www.rfc-editor.org/rfc/rfc8832.html#name-data_channel_ack-message): reception of the ACK tells the opener the handshake is complete.

## Verification

```bash
cd rtc/
cargo test -p rtc-datachannel --lib
cargo test --lib create_data_channel_dials_immediately_when_sctp_association_present
cargo clippy -- -D warnings
```

### DIFF
--- a/rtc-datachannel/src/data_channel/mod.rs
+++ b/rtc-datachannel/src/data_channel/mod.rs
@@ -78,6 +78,9 @@ pub struct DataChannel {
     read_outs: VecDeque<DataChannelMessage>,
     write_outs: VecDeque<DataChannelMessage>,
 
+    /// Whether the DCEP handshake for this channel has completed.
+    handshake_complete: bool,
+
     // stats
     messages_sent: usize,
     messages_received: usize,
@@ -129,6 +132,11 @@ impl DataChannel {
             negotiated: config.negotiated,
         });
 
+        // Out-of-band `negotiated` channels have no handshake, so they are
+        // handshake-complete immediately. In-band channels stay incomplete until
+        // `handle_dcep` processes the peer's `DATA_CHANNEL_ACK`.
+        data_channel.handshake_complete = config.negotiated;
+
         Ok(data_channel)
     }
 
@@ -160,6 +168,10 @@ impl DataChannel {
         let mut data_channel = DataChannel::new(config, association_handle, stream_id);
 
         data_channel.write_data_channel_ack()?;
+
+        // The remote side's OPEN has been processed and the ACK is written out,
+        // so the handshake is complete.
+        data_channel.handshake_complete = true;
 
         Ok(data_channel)
     }
@@ -199,6 +211,15 @@ impl DataChannel {
         &self.config
     }
 
+    /// Whether the DCEP handshake for this channel has completed.
+    ///
+    /// For a locally-created in-band channel this becomes `true` only when the peer's
+    /// `DATA_CHANNEL_ACK` is received; for a remote channel it is `true` from `accept()`;
+    /// for an out-of-band `negotiated` channel it is `true` from `dial()`.
+    pub fn is_handshake_complete(&self) -> bool {
+        self.handshake_complete
+    }
+
     fn handle_dcep<B>(&mut self, data: &mut B) -> Result<()>
     where
         B: Buf,
@@ -214,6 +235,8 @@ impl DataChannel {
             }
             Message::DataChannelAck(_) => {
                 debug!("Received DATA_CHANNEL_ACK");
+                // The initiator's DCEP handshake is complete.
+                self.handshake_complete = true;
             }
             _ => {
                 return Err(Error::InvalidMessageType(msg.message_type() as u8));
@@ -442,6 +465,8 @@ impl sansio::Protocol<DataChannelMessage, DataChannelMessage, ()> for DataChanne
         // a corresponding notification to the application layer that the reset
         // has been performed.  Streams are available for reuse after a reset
         // has been performed.
+        // Closing resets the handshake-complete flag.
+        self.handshake_complete = false;
         self.write_data_channel_close()
     }
 }
--- a/src/peer_connection/handler/datachannel.rs
+++ b/src/peer_connection/handler/datachannel.rs
@@ -15,7 +15,7 @@ use sctp::PayloadProtocolIdentifier;
 use shared::TransportContext;
 use shared::error::{Error, Result};
 use std::collections::{HashMap, VecDeque};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 pub(crate) struct DataChannelHandlerContext {
     pub(crate) read_outs: VecDeque<TaggedRTCMessageInternal>,
@@ -53,6 +53,8 @@ pub(crate) struct DataChannelHandler<'a> {
     ctx: &'a mut DataChannelHandlerContext,
     data_channels: &'a mut HashMap<RTCDataChannelId, RTCDataChannelInternal>,
     stats: &'a mut RTCStatsAccumulator,
+    /// Configured DCEP handshake timeout for in-band channels. `None` disables it.
+    dcep_handshake_timeout: Option<Duration>,
 }
 
 impl<'a> DataChannelHandler<'a> {
@@ -60,16 +62,49 @@ impl<'a> DataChannelHandler<'a> {
         ctx: &'a mut DataChannelHandlerContext,
         data_channels: &'a mut HashMap<RTCDataChannelId, RTCDataChannelInternal>,
         stats: &'a mut RTCStatsAccumulator,
+        dcep_handshake_timeout: Option<Duration>,
     ) -> Self {
         DataChannelHandler {
             ctx,
             data_channels,
             stats,
+            dcep_handshake_timeout,
         }
     }
 
     pub(crate) fn name(&self) -> &'static str {
         "DataChannelHandler"
+    }
+
+    /// Emit the `DataChannelEvent::Open` application message and record the
+    /// corresponding peer-connection and per-channel statistics.
+    ///
+    /// The caller must have already ensured the channel's `ready_state` is `Open`.
+    fn emit_data_channel_opened(
+        &mut self,
+        now: Instant,
+        transport: TransportContext,
+        id: RTCDataChannelId,
+    ) -> Result<()> {
+        let dc = self
+            .data_channels
+            .get(&id)
+            .ok_or(Error::ErrDataChannelNotExisted)?;
+
+        self.ctx.read_outs.push_back(TaggedRTCMessageInternal {
+            now,
+            transport,
+            message: RTCMessageInternal::Dtls(DTLSMessage::DataChannel(ApplicationMessage {
+                data_channel_id: id,
+                data_channel_event: DataChannelEvent::Open,
+            })),
+        });
+
+        self.stats.peer_connection.on_data_channel_opened();
+        self.stats
+            .get_or_create_data_channel(id, &dc.label, &dc.protocol)
+            .on_state_changed(RTCDataChannelState::Open);
+        Ok(())
     }
 }
 
@@ -94,13 +129,32 @@ impl<'a>
             );
 
             let stream_id = message.stream_id;
+            let transport = msg.transport;
 
-            if let Some(data_channel_internal) = self.data_channels.get_mut(&stream_id) {
+            let opened = if let Some(data_channel_internal) = self.data_channels.get_mut(&stream_id)
+            {
+                // A closed channel is terminal: ignore any late DCEP or user data.
+                if data_channel_internal.ready_state == RTCDataChannelState::Closed {
+                    return Ok(());
+                }
+
+                // Process the inbound message, then check whether a still-connecting
+                // in-band channel just completed its DCEP handshake (initiator side).
+                // If so, promote it to Open and emit the open event.
+                let mut opened = false;
                 let data_channel = data_channel_internal
                     .data_channel
                     .as_mut()
                     .ok_or(Error::ErrDataChannelNotExisted)?;
                 data_channel.handle_read(message)?;
+                if data_channel.is_handshake_complete()
+                    && data_channel_internal.ready_state == RTCDataChannelState::Connecting
+                {
+                    data_channel_internal.ready_state = RTCDataChannelState::Open;
+                    data_channel_internal.handshake_deadline = None;
+                    opened = true;
+                }
+                opened
             } else {
                 let data_channel_internal = RTCDataChannelInternal::accept(
                     message.association_handle,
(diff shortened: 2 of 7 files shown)


dimitri-lenkov:
Force pushed with rustfmt import reordering

https://github.com/webrtc-rs/rtc/compare/385162c141cdb069bcc71ea59806f7ec1a58a04c..38aebd50cd7ebcc8ab814740e4fcba87a78246ad

sorry I'm used to not changing import orders usually, hah.

dimitri-lenkov:
Updated branch to latest changes, and updated OP title & text so it's easier to understand (removed the pion references).

src/peer_connection/handler/datachannel.rs
```diff
@@ -71,6 +71,37 @@ impl<'a> DataChannelHandler<'a> {
...
+    /// The caller must have already ensured the channel exists and that its
+    /// `ready_state` is `Open`.
+    fn emit_data_channel_opened(
+        &mut self,
+        now: Instant,
+        transport: TransportContext,
+        id: RTCDataChannelId,
+    ) {
+        let (label, protocol) = {
+            let dc = self.data_channels.get(&id).expect("channel must exist");
```
rainliu, line 87: never use .expect, instead, return Err(Error::Other("channel must exist").

so, emit_data_channel_opened return Result<()>

src/peer_connection/handler/datachannel.rs
```diff
@@ -71,6 +71,37 @@ impl<'a> DataChannelHandler<'a> {
...
+            (dc.label.clone(), dc.protocol.clone())
```
rainliu, line 88: can we avoid .clone()?

rtc-datachannel/src/data_channel/mod.rs
```diff
@@ -78,6 +78,9 @@ pub struct DataChannel {
     read_outs: VecDeque<DataChannelMessage>,
     write_outs: VecDeque<DataChannelMessage>,
 
+    /// Whether the DCEP handshake for this channel has completed.
+    open: bool,
```
rainliu, line 82: open is misleading, why not just use `pub negotiated: bool,`

rtc-datachannel/src/data_channel/mod.rs
```diff
@@ -199,6 +211,15 @@ impl DataChannel {
         &self.config
     }
 
+    /// Whether the DCEP handshake for this channel has completed.
+    ///
+    /// For a locally-created in-band channel this becomes `true` only when the peer's
+    /// `DATA_CHANNEL_ACK` is received; for a remote channel it is `true` from `accept()`;
+    /// for an out-of-band `negotiated` channel it is `true` from `dial()`.
+    pub fn is_open(&self) -> bool {
```
rainliu, line 219: follow above, rename to `is_negotiated()`

rainliu (changes requested):

src/peer_connection/handler/datachannel.rs
rainliu:
```
    ) -> Result<()> {
        let dc = self
            .data_channels
            .get(&id)
            .ok_or(Error::ErrDataChannelNotExisted)?;
```

src/peer_connection/handler/datachannel.rs
```diff
@@ -71,6 +71,37 @@ impl<'a> DataChannelHandler<'a> {
...
+            transport,
+            message: RTCMessageInternal::Dtls(DTLSMessage::DataChannel(ApplicationMessage {
+                data_channel_id: id,
+                data_channel_event: DataChannelEvent::Open,
+            })),
+        });
+
+        self.stats.peer_connection.on_data_channel_opened();
+        self.stats
+            .get_or_create_data_channel(id, &label, &protocol)
```
rainliu, line 102:
```
 self.stats
            .get_or_create_data_channel(id, &dc.label, &dc.protocol)
            .on_state_changed(RTCDataChannelState::Open);
```

src/peer_connection/handler/datachannel.rs
```diff
@@ -296,36 +323,23 @@ impl<'a>
         let now = evt.now;
         match evt.event {
             RTCEventInternal::SCTPHandshakeComplete(association_handle) => {
+                // Out-of-band negotiated channels have no DCEP handshake, so they are
+                // open immediately and fire the open event here. In-band channels stay
+                // connecting until their `DATA_CHANNEL_ACK` arrives in `handle_read`.
+                let mut opened = Vec::new();
                 for data_channel_internal in self.data_channels.values_mut() {
                     if data_channel_internal.ready_state == RTCDataChannelState::Connecting {
```
rainliu, line 331: Before this PR that was a sufficient guard, because dial() set ready_state = Open. Now an in-band channel stays Connecting after dialing, so the guard no longer excludes it.

Fix is one line, matching the guard create_data_channel already uses:
```
if data_channel_internal.ready_state == RTCDataChannelState::Connecting
    && data_channel_internal.data_channel.is_none()
{
```

rtc-datachannel/src/data_channel/mod.rs
rainliu: or rename to `handshake_complete: bool`

and below
`fn  is_handshake_complete()`

rtc-datachannel/src/data_channel/mod.rs
```diff
@@ -214,6 +235,8 @@ impl DataChannel {
             }
             Message::DataChannelAck(_) => {
                 debug!("Received DATA_CHANNEL_ACK");
+                // The initiator's DCEP handshake is complete.
+                self.open = true;
```
rainliu, line 239: what if a DATA_CHANNEL_ACK arriving on a closed channel?

it should  be gated on RTCDataChannelState::Connecting

src/peer_connection/handler/datachannel.rs
rainliu, line 408: no timeout if the ACK never arrives. With promotion happening only on the ACK, a channel whose ACK is lost sits in Connecting forever: no OnOpen, data_channels_opened stays 0 while a later SCTPStreamClosed still increments data_channels_closed, and DataChannelEvent::Message gets delivered for a channel the app was never told about.

dimitri-lenkov:
Thanks for the review. GitHub was down but I got your responses via email and started work.

Force pushed the latest changes. I'm about to run off to work but when I'm back (or tomorrow morning) I can address any other additional feedback.

All changes in relation to the review notes is below, in the order raised.

---

## 1. `emit_data_channel_opened` must return `Result<()>`, no `.expect`

> never use .expect, instead, return Err(Error::Other("channel must exist"). so, emit_data_channel_opened return Result<()>

`emit_data_channel_opened` now returns `Result<()>`, and the `.expect` is gone. The channel is fetched with `Error::ErrDataChannelNotExisted` (rather than `Error::Other("channel must exist")`), because that variant is already used for every missing-channel case elsewhere in this handler.

```rust
    fn emit_data_channel_opened(
        &mut self,
        now: Instant,
        transport: TransportContext,
        id: RTCDataChannelId,
    ) -> Result<()> {
        let dc = self
            .data_channels
            .get(&id)
            .ok_or(Error::ErrDataChannelNotExisted)?;
        // ...
        Ok(())
    }
```

Both call sites propagate the error with `?`:
- `handle_read` after the ACK promotes the channel.
- `handle_event` for negotiated channels at `SCTPHandshakeComplete`.

## 2. Avoid cloning `label` / `protocol`

> can we avoid .clone()?

The `emit_data_channel_opened` helper no longer clones; the stats call is fed the borrowed channel directly:

```rust
        self.stats
            .get_or_create_data_channel(id, &dc.label, &dc.protocol)
            .on_state_changed(RTCDataChannelState::Open);
```

## 3. Rename the `open` field

> open is misleading, why not just use `pub negotiated: bool,`
>
> or rename to `handshake_complete: bool`
>
> and below `fn is_handshake_complete()`

`handshake_complete` was chosen, following the second suggestion. It covers both cases: negotiated channels are handshake-complete immediately, and in-band channels become handshake-complete once the peer's `DATA_CHANNEL_ACK` is processed.

```rust
pub struct DataChannel {
    // ...
    handshake_complete: bool,
}
```

- `dial()` sets `handshake_complete = config.negotiated`.
- `accept()` sets `handshake_complete = true`.
- `handle_dcep` `DataChannelAck` sets `handshake_complete = true`.
- `close()` sets `handshake_complete = false`.

## 4. Rename `is_open()` to `is_handshake_complete()`

> follow above, rename to `is_negotiated()` (elsewhere) `fn is_handshake_complete()`

Renamed for the same reason as the field:

```rust
    pub fn is_handshake_complete(&self) -> bool {
        self.handshake_complete
    }
```

All callers and tests were updated, including the data-channel crate's unit tests (e.g. `test_data_channel_close_resets_handshake_complete_flag`).

## 5. `SCTPHandshakeComplete` must not re-dial already-dialed channels

> Before this PR that was a sufficient guard, because dial() set ready_state = Open. Now an in-band channel stays Connecting after dialing, so the guard no longer excludes it.
>
> Fix is one line, matching the guard create_data_channel already uses:
> ```rust
> if data_channel_internal.ready_state == RTCDataChannelState::Connecting
>     && data_channel_internal.data_channel.is_none()
> {
> ```

Applied as suggested:

```rust
if data_channel_internal.ready_state == RTCDataChannelState::Connecting
    && data_channel_internal.data_channel.is_none()
{
    data_channel_internal.dial(association_handle)?;
    // ...
}
```

## 6. `DATA_CHANNEL_ACK` on a closed channel

> what if a DATA_CHANNEL_ACK arriving on a closed channel? it should be gated on RTCDataChannelState::Connecting

A closed channel is terminal. `handle_read` now returns early before calling into the datachannel layer when the channel is already `Closed`, so a straggler ACK (or any late DCEP/user data) cannot mutate it:

```rust
if data_channel_internal.ready_state == RTCDataChannelState::Closed {
    return Ok(());
}
```

## 7. No timeout if the ACK never arrives

> no timeout if the ACK never arrives. With promotion happening only on the ACK, a channel whose ACK is lost sits in Connecting forever: no OnOpen, data_channels_opened stays 0 while a later SCTPStreamClosed still increments data_channels_closed, and DataChannelEvent::Message gets delivered for a channel the app was never told about.

Implemented a full timeout mechanism.

**Configurable setting** - `dcep_handshake_timeout: Option<Duration>` is exposed on a `data_channel` config sub-struct of `SettingEngine` (alongside `candidates`, `multicast_dns`, and `replay_protection`), set via `SettingEngineBuilder::with_dcep_handshake_timeout`. `Some(...)` arms the timeout; `None` disables it. It defaults to `Some(30s)`. I wasn't sure what the default should be, so I set it to 30s which felt like a good default.

**Per-channel bookkeeping** - `RTCDataChannelInternal` holds `handshake_deadline: Option<Instant>`. The deadline is armed when an in-band channel is dialed, and cleared on promotion or close.

**Timeout handling** - `poll_timeout` reports the earliest deadline among `Connecting` channels; `handle_timeout` closes expired channels: it sets `ready_state = Closed`, closes the underlying channel, records open/closed stats, marks `close_emitted = true`, and emits `OnClose`.

```rust
for id in timed_out {
    if let Some(dc) = self.data_channels.get_mut(&id) {
        dc.handshake_deadline = None;
        dc.ready_state = RTCDataChannelState::Closed;
        if let Some(data_channel) = dc.data_channel.as_mut() {
            data_channel.close()?;
        }

        self.stats.peer_connection.on_data_channel_closed();
        self.stats
            .get_or_create_data_channel(id, &dc.label, &dc.protocol)
            .on_state_changed(RTCDataChannelState::Closed);

        dc.close_emitted = true;
        self.ctx.event_outs.push_back(TaggedRTCEventInternal {
            now,
            event: RTCEventInternal::RTCPeerConnectionEvent(
                RTCPeerConnectionEvent::OnDataChannel(RTCDataChannelEvent::OnClose(id)),
            ),
        });
    }
}
```

**Avoiding double-close** - `SCTPStreamClosed` emits `OnClose` and counts the close only when `!dc.close_emitted`. `close_emitted` is set **only** by the timeout path, so "closed by timeout" is distinct from "closed normally" (a normally-closed channel also has `ready_state == Closed`, so a `ready_state` check would suppress the legitimate `OnClose`). This is exercised by the existing integration test `data_channels_close_by_rtc_interop`.

**Buffering pre-open data** - application messages are delivered only when the channel's `ready_state == Open`. Messages arriving while the channel is still `Connecting` stay buffered in the underlying `DataChannel` read queue; they drain after the ACK promotes the channel, or are discarded on timeout/close. This addresses the "Message gets delivered for a channel the app was never told about" part of your comment.

---

## Added tests

All new behavior is pinned by unit tests in `src/peer_connection/handler/datachannel.rs`:

- `emit_data_channel_opened_missing_channel_returns_error` - missing channel returns `Err(Error::ErrDataChannelNotExisted)` instead of panicking.
- `sctp_handshake_complete_does_not_redial` - a second `SCTPHandshakeComplete` queues no new `DATA_CHANNEL_OPEN`.
- `negotiated_channel_fires_open_at_handshake_complete` - a negotiated channel opens and fires the open event immediately at `SCTPHandshakeComplete`, with no deadline armed.
- `ack_on_closed_channel_is_ignored` - a late ACK on a `Closed` channel is ignored.
- `in_band_channel_times_out_without_ack` - an in-band channel whose ACK never arrives transitions to `Closed`, fires `OnClose` once, and is counted exactly once.
- `pre_open_data_is_buffered_until_open` - data before the ACK is withheld and delivered after the open event.
- `pre_open_data_is_dropped_on_timeout` - buffered data is never delivered when the channel times out.

rainliu (approved):

PR #182, opened 2026-08-17
Bind declared remote streams after codec discovery

ri-nat:
## Summary

- retry remote interceptor binding when the first RTP packet resolves an SDP-declared SSRC's deferred codec
- bind only that exact coding and its RTX/FEC repair streams, avoiding state resets on already-active simulcast streams
- add regressions proving rtc-generated TWCC resumes and each declared simulcast/repair SSRC binds exactly once

## Root cause

`start_rtp` creates SDP-declared tracks with an empty codec and immediately attempts interceptor binding. The codec fuzzy match correctly skips that incomplete coding. The endpoint later resolves and stores the codec from the first RTP payload type, but previously never retried the remote-stream bind, leaving receiver-report, NACK, and TWCC interceptors unaware of the stream.

## Validation

- `cargo fmt --all -- --check`
- `cargo test --lib declared_ -- --nocapture`
- `cargo clippy -p rtc --lib --tests --no-deps`
- `cargo test -p rtc --tests -- --skip test_ice_restart_interop`
- `cargo test --workspace --exclude rtc --exclude rtc-mdns --exclude rtc-ice`
- `cargo test -p rtc-mdns -p rtc-ice`

The unskipped `ice_restart_by_webrtc_interop` test is unrelated to this change and fails on this macOS host while the legacy `webrtc` peer enumerates unavailable link-local IPv6 interfaces (`Can't assign requested address`).

### DIFF
--- a/src/peer_connection/handler/endpoint.rs
+++ b/src/peer_connection/handler/endpoint.rs
@@ -408,6 +408,17 @@ where
                     let new_entry = receiver.track_mut().set_codec_by_ssrc(codec, ssrc);
                     assert!(!new_entry);
 
+                    // The eager bind performed while applying the remote SDP
+                    // skipped this declared SSRC because its codec was deferred.
+                    // Bind this exact coding now, before OnTrack is emitted;
+                    // the empty-codec guard above makes this once-only.
+                    receiver.interceptor_remote_stream_by_ssrc_op(
+                        self.media_engine,
+                        self.interceptor,
+                        true,
+                        ssrc,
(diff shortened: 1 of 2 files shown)


ri-nat:
I'm really sorry for my agent being too optimistic. The bug is real, but I don't want to push my vibe-coded solution to the main repo. Sorry once again.

rainliu:
@ri-nat , could you file an issue to describe this bug and how to reproduce it? Thanks.

Issue #178, opened 2026-08-17
Default SCTP packet sizing overflows the 1280-byte IPv6 minimum MTU: datachannel messages silently blackholed on tunnel paths (regression of webrtc#807)


lovon-spec:
## Summary

Two compounding problems make `rtc-sctp` emit UDP datagrams that exceed the 1280-byte IPv6 minimum MTU:

1. **The default is too large.** `INITIAL_MTU` is 1228, so even a correctly-sized packet becomes ~1313 wire bytes after DTLS and IPv6/UDP overhead.
2. **The MTU isn't enforced exactly.** `bundle_data_chunks_into_packets` decides bundle boundaries on unpadded payload sizes and omits the candidate chunk's 16-byte DATA header, so bundles the check accepts can marshal past whatever MTU is configured. The fast-retransmit sizing loop has the same unpadded accounting, and a single maximum-size chunk marshals to `INITIAL_MTU + 1` after final-chunk padding.

On paths whose MTU is exactly 1280 — Tailscale's default, and a common conservative setting for WireGuard and other UDP tunnels — the oversized datagram is dropped (IPv6 forbids on-path fragmentation; RFC 8900 discourages relying on it anywhere), and the loss is **silent and permanent for the life of the association**: retransmissions re-bundle at the same size, `rtc-sctp` performs no path-MTU probing, `send` returns `Ok`, and no error ever surfaces — while RTP media on the same candidate pair keeps flowing normally.

Problem 1 is a regression of the fix the async `webrtc` crate shipped in v0.17.2 (webrtc-rs/webrtc#806 → webrtc-rs/webrtc#807, following pion/sctp#476): the 0.20 architecture replaced that sctp implementation with `rtc-sctp`, which still carries 1228. There is also no configuration escape hatch: the peer connection constructs `EndpointConfig::default()` unconditionally and `SettingEngineBuilder` exposes no knob that reaches it.

## Wire math

Each assembled SCTP packet becomes one DTLS application-data record carried in one UDP datagram:

```
1228 SCTP packet
+ ~37 DTLS 1.2 record overhead (13-byte header + 8-byte explicit nonce + 16-byte AEAD tag, AES-GCM)
+  8 UDP header
+ 40 IPv6 header
= ~1313 wire bytes  >  1280 IPv6 minimum MTU
```

1191 is the exact worst-case budget for this stack, including a TURN relay — the derivation pion/sctp#476 established and webrtc-rs/webrtc#807 adopted:

```
[...]

PR #184, opened 2026-08-17
Add SettingEngineBuilder::with_sctp_mtu for non-default path MTUs

lovon-spec:
Follow-up promised in #180: a send-MTU knob for paths whose budget differs from the default (1191, the TURN-relayed IPv6 budget — #178).

`SettingEngineBuilder::with_sctp_mtu(mtu)` converts the requested DATA-packet budget with rtc-sctp's new `max_payload_size_for_mtu` helper, then applies it through the existing `EndpointConfig::max_payload_size` setter:

- Derives `max_payload_size` as the exact inverse of the default derivation — subtract the 12-byte common and 16-byte DATA chunk headers, round down to the 4-byte padding boundary — so a single maximum-size DATA chunk marshals within the budget, and #180's padded-wire bundling keeps multi-chunk DATA packets inside it.
- Scope: an outbound **DATA-packet** budget. Control packets (INIT/INIT ACK, SACK, RECONFIG, FORWARD TSN, shutdown) marshal without consulting it.
- Inputs below 32 — the smallest representable padded DATA packet — are raised to 32 with a warning.

Also fixed here: `Association::new` initialized the congestion window as `(2 * mtu).clamp(4380, 4 * mtu)`, which panics for effective MTUs below ~1095 (`Ord::clamp` with min > max) and overflows near `u32::MAX`. Latent behind the low-level `max_payload_size` setter, but this knob advertises exactly such MTUs — rewritten in RFC 4960 §7.2.1's total form with saturating arithmetic.

Tests: the derivation (default-equivalence, a rounding boundary, the 32 floor, `u32::MAX`); emitted single-chunk packet sizes at the floor, across a padding boundary, and at 1500 (reusing #180's helpers); start-time plumbing; and a cwnd construction regression (small and `u32::MAX`).

Assisted-by: claude:claude-fable-5

### DIFF
--- a/rtc-sctp/src/config.rs
+++ b/rtc-sctp/src/config.rs
@@ -21,6 +21,10 @@ pub(crate) const INITIAL_MTU: u32 = 1191;
 pub(crate) const INITIAL_RECV_BUF_SIZE: u32 = 1024 * 1024;
 pub(crate) const COMMON_HEADER_SIZE: u32 = 12;
 pub(crate) const DATA_CHUNK_HEADER_SIZE: u32 = 16;
+/// Floor for [`EndpointConfig::mtu`]: the smallest representable padded DATA
+/// packet — the 12-byte common header, the 16-byte DATA chunk header, and a
+/// 1..=4-byte payload padded to the 4-byte chunk boundary.
+pub(crate) const MIN_DATA_PACKET_MTU: u32 = COMMON_HEADER_SIZE + DATA_CHUNK_HEADER_SIZE + 4;
 pub(crate) const DEFAULT_MAX_MESSAGE_SIZE: u32 = 262144;
 
 /// Config collects the arguments to create_association construction into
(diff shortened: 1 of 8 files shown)


rtc-sctp/src/config.rs
```diff
@@ -181,6 +185,35 @@ impl EndpointConfig {
...
+    /// SCTP 4-byte chunk-padding boundary —
+    /// `(value - COMMON_HEADER_SIZE - DATA_CHUNK_HEADER_SIZE) & !3` — so a single maximum-size
+    /// DATA chunk (common header plus the padded chunk) always marshals to at most `value`
+    /// bytes, exactly as the default budget does for `INITIAL_MTU` (1191, the TURN-relayed
+    /// IPv6 minimum-MTU budget).
+    ///
+    /// Values below 32 — the smallest representable padded DATA packet: the two headers
+    /// plus a 1..=4-byte payload padded to the 4-byte boundary — cannot satisfy that
+    /// promise and are raised to 32 with a warning.
+    pub fn mtu(&mut self, value: u32) -> &mut Self {
```
rainliu: I think such API design may have issue.

what if rtc-sctp user first call mtu(), then call max_payload_size?
```
EndpointConfig.mtu(1000).max_payload_size(2000)?
```

rainliu (changes requested):

rtc-sctp/src/config.rs
lovon-spec: Good catch, agreed. Two setters writing the same field like that is confusing. I reworked it in a follow-up commit: dropped the `mtu()` setter and made the derivation a plain function (`max_payload_size_for_mtu`) that the rtc side calls before `max_payload_size()`. That way `EndpointConfig` keeps a single setter and there's no call order to get wrong. The floor, warning, docs and tests moved over unchanged.

rainliu (approved):

rainliu (approved):

PR #200, opened 2026-08-25
Wait for streams to drain before reset

### LINKED ISSUE #816: [v0.17.x] sctp: incoming stream reset discards data already delivered to the reassembly queue
`Stream::read_sctp` checks `read_shutdown` before draining the reassembly queue:

```rust
// sctp/src/stream/mod.rs (v0.17.x, line 182; identical in webrtc-sctp 0.13.0)
pub async fn read_sctp(&self, p: &mut [u8]) -> Result<(usize, PayloadProtocolIdentifier)> {
    loop {
        if self.read_shutdown.load(Ordering::SeqCst) {
            return Ok((0, PayloadProtocolIdentifier::Unknown));
        }

        let result = {
            let mut reassembly_queue = self.reassembly_queue.lock().await;
            reassembly_queue.read(p)
        };
        ...
```

When an incoming stream reset arrives, `reset_streams_if_any` → `unregister_stream` (`sctp/src/association/association_internal.rs:328`) sets `read_shutdown` immediately. Per RFC 6525 the reset is only performed once `peer_last_tsn` reaches the request's `sender_last_tsn` — i.e. **only after all of the stream's data has been received into the reassembly queue** (the `sna32lte` gate in `reset_streams_if_any` enforces this correctly). So at the moment `read_shutdown` is set, the queue can still hold data that was received in full and in order — and the next `read` returns `Ok(0)` (EOF) without delivering it.

**Consequence:** if a peer writes to a data channel and then promptly closes it (`send(...)` → `close()` — a very common pattern, e.g. any request/response or echo exchange), the receiver races its read loop against the reset and silently loses the data. Against a pion echo server we reproduce ~10% data loss on loopback; the wire trace shows everything delivered correctly — the loss is purely local.

**Sequence:**

1. Peer: `DATA` chunks → `close()` → outgoing reset request
2. Receiver association: DATA chunks land in the reassembly queue (readable)
3. Receiver association: reconfig processed (TSN gate passes) → `unregister_stream` → `read_shutdown = true`
4. Receiver's data-channel read loop wakes, calls `read_sctp` → hits the `read_shutdown` early-return → `Ok(0)` (EOF), queued data never delivered

Whether the read loop wins the race at step 4 is timing-dependent, hence intermittent.

**pion/sctp** (which this crate mirrors) reads the queue first and only reports the terminal state once the queue is empty:

```go
// pion/sctp stream.go ReadSCTP
for {
    n, ppi, err := s.reassemblyQueue.read(payload)
    if err == nil || errors.Is(err, io.ErrShortBuffer) {
        return n, ppi, err
    }
    if s.readErr != nil {
        return 0, PayloadProtocolIdentifier(0), s.readErr
    }
    s.readNotifier.Wait()
}
```

**Suggested fix** — drain before honoring shutdown, and register the notifier before the state check so a shutdown between the queue read and the await can't be lost (`notify_waiters` doesn't store a permit):

```rust
pub async fn read_sctp(&self, p: &mut [u8]) -> Result<(usize, PayloadProtocolIdentifier)> {
    loop {
        let notified = self.read_notifier.notified();

        let result = {
            let mut reassembly_queue = self.reassembly_queue.lock().await;
            reassembly_queue.read(p)
        };

        match result {
            Ok(_) | Err(Error::ErrShortBuffer { .. }) => return result,
            Err(_) => {
                if self.read_shutdown.load(Ordering::SeqCst) {
                    return Ok((0, PayloadProtocolIdentifier::Unknown));
                }
                notified.await;
            }
        }
    }
}
```

We've been running this patch against pion and libdatachannel peers: the ~10% loss goes to 0/40 on the repro, with no other behavior change observed. Patched source for reference: https://github.com/privacy-ethereum/kps/commit/2d730447917bbf775cda0ad967ea3e737de46ef7 (verbatim webrtc-sctp 0.13.0 + this one change).

**Affected:** `webrtc-sctp` 0.13.0 (via `webrtc` 0.14.0); code is unchanged on the `v0.17.x` branch. I haven't audited the sans-IO rewrite on `master`.

Happy to open a PR against `v0.17.x` if that's welcome.

voltrevo:
Fixes https://github.com/webrtc-rs/webrtc/issues/816.

PR per request, since the issue was still present in 0.20+.

> If not, could you submit a PR to master branch too?

### DIFF
--- a/rtc-sctp/src/association/mod.rs
+++ b/rtc-sctp/src/association/mod.rs
@@ -137,9 +137,9 @@ pub struct Association {
     bytes_received: usize,
     bytes_sent: usize,
 
-    peer_verification_tag: u32,
+    pub(crate) peer_verification_tag: u32,
     my_verification_tag: u32,
-    my_next_tsn: u32,
+    pub(crate) my_next_tsn: u32,
     peer_last_tsn: u32,
     // for RTT measurement
     min_tsn2measure_rtt: u32,
@@ -166,13 +166,13 @@ pub struct Association {
     local_addr: SocketAddr,
     transport_protocol: TransportProtocol,
 
-    source_port: u16,
-    destination_port: u16,
+    pub(crate) source_port: u16,
+    pub(crate) destination_port: u16,
     my_max_num_inbound_streams: u16,
     my_max_num_outbound_streams: u16,
     my_cookie: Option<ParamStateCookie>,
 
-    payload_queue: PayloadQueue,
+    pub(crate) payload_queue: PayloadQueue,
     inflight_queue: PayloadQueue,
     pending_queue: PendingQueue,
     control_queue: VecDeque<Packet>,
@@ -787,6 +787,36 @@ impl Association {
             });
             s.state = RecvSendState::Closed;
         }
+    }
+
+    /// Re-run any reset request that was answered In Progress because
+    /// `stream_identifier` still had data the application had not collected
+    /// Called when that stream drains.
+    ///
+    /// The reply is queued like any other control packet; a request naming
+    /// several streams is retried as a whole and simply defers again if another
+    /// of its streams is still readable.
+    pub(crate) fn retry_deferred_resets(&mut self, stream_identifier: StreamId) {
+        let pending: Vec<ParamOutgoingResetRequest> = self
+            .reconfig_requests
+            .values()
+            .filter(|p| p.stream_identifiers.contains(&stream_identifier))
+            .cloned()
+            .collect();
+        if pending.is_empty() {
+            return;
+        }
+
+        let mut reply = vec![];
+        for p in pending {
+            if let Err(err) = self.reset_streams_if_any(&p, true, &mut reply) {
+                debug!("[{}] retry of deferred reset failed: {:?}", self.side, err);
+            }
+        }
+        for packet in reply {
+            self.control_queue.push_back(packet);
+        }
+        self.awake_write_loop();
     }
 
     /// set_state atomically sets the state of the Association.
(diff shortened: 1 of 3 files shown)


rtc-sctp/src/association/mod.rs
```diff
@@ -137,9 +137,9 @@ pub struct Association {
     bytes_received: usize,
     bytes_sent: usize,
```
voltrevo, line 138: pub(crate) applied below for use in new tests.

rtc-sctp/src/association/mod.rs
```diff
@@ -1977,16 +2025,31 @@ impl Association {
                 "[{}] resetStream(): senderLastTSN={} <= peer_last_tsn={}",
                 self.side, p.sender_last_tsn, self.peer_last_tsn
             );
+
```
voltrevo, line 2028: An incoming reset is only processed once everything up to `sender_last_tsn` has arrived, so by construction the resetting peer's data is already in the reassembly queue. Performing the reset now discards whatever the application has not collected yet — a peer that writes and immediately closes loses the write.

RFC 6525 §5.2.2 already has the answer for "cannot perform this yet": respond In Progress and keep the request. Only a *readable* message holds it back; an incomplete fragment set can never be completed once the peer has reset, so it is genuinely discardable and must not stall the reset forever. `Stream::read_sctp` retries the request as soon as the queue runs dry, so this costs no extra round trip in the common case.

rtc-sctp/src/association/stream.rs
```diff
@@ -127,7 +127,12 @@ impl Stream<'_> {
         if let Some(s) = self.association.streams.get_mut(&self.stream_identifier)
             && (s.state == RecvSendState::ReadWritable || s.state == RecvSendState::Readable)
         {
-            Ok(s.reassembly_queue.read())
+            let chunks = s.reassembly_queue.read();
```
voltrevo, line 130: A reset request this stream was holding back (see `reset_streams_if_any`) becomes performable the moment the application has taken everything. Retrying here rather than waiting for the peer to retransmit its RECONFIG is what keeps the common write-then-close path free of an extra round trip.

rainliu (approved):

rainliu:
please fix fmt/clippy

voltrevo:
done 🙏

PR #201, opened 2026-08-25
Fix deadlock in buffer that has a gap but is full

### LINKED ISSUE #822: sctp: permanent zero-window deadlock when the chunk dropped at buffer-full is the tail of a burst
**Affected:** `webrtc-sctp` 0.13.0; the relevant code is unchanged through
v0.17.1 (`sctp/src/association/association_internal.rs::handle_data`).

### Summary

`handle_data`'s buffer-full path can permanently deadlock an association: the
receiver ends up at a zero window it can never reopen, and the sender
retransmits the same chunk forever with every copy dropped.

The buffer-full branch only accepts a chunk if the payload queue is non-empty
and the chunk fills a gap below the highest received TSN:

```rust
} else {
    // Receive buffer is full
    if let Some(last_tsn) = self.payload_queue.get_last_tsn_received() {
        if sna32lt(d.tsn, *last_tsn) {
            // accepted as a missing chunk
            ...
        }
    } else {
        // dropped
        ...
    }
}
```

Failure sequence:

1. A sender bursts enough data to cross `max_receive_buffer_size` while the
   application hasn't drained yet; the chunk that crosses the line is dropped
   ("receive buffer full"). Suppose that chunk is the **last** outstanding one
   of the burst (e.g. the final fragment of a fragmented user message).
2. Everything received before it is in sequence, so `peer_last_tsn` advances
   over the whole payload queue and it drains to empty.
3. The dropped chunk's earlier fragments sit in the reassembly queue as an
   incomplete chunk set. Messages are reassembled whole and delivered in
   order, so nothing in the queue is deliverable (complete messages behind
   the incomplete head-of-line set are blocked too): `read` keeps returning
   the `ErrTryAgain` would-block sentinel, the bytes are never subtracted,
   and `get_my_receiver_window_credit()` stays at 0.
4. The sender's T3-rtx retransmits the missing chunk — but now
   `get_last_tsn_received()` is `None`, so the retransmission takes the
   unconditional drop branch. Every subsequent retransmission is dropped the
   same way, forever.

Nothing can break the cycle: the application cannot drain (incomplete set),
the window cannot reopen (bytes pinned), and the only chunk that would fix
both is dropped on every arrival.

Observed as an intermittent (~5%) permanent stall of pion → webrtc-rs
transfers of ~1.5 MB, whenever a burst crossed the 1 MB default receive
buffer at a message-fragment boundary near the burst tail.

### Reproduction

Minimal standalone crate (attached: [`repro-sctp-zero-window.zip`](https://github.com/user-attachments/files/30179491/repro-sctp-zero-window.zip)), fully
deterministic — no WebRTC, just two `webrtc-sctp` associations over loopback
UDP. A single 48 KiB user message to a receiver with a 32 KiB
`max_receive_buffer_size` never delivers and never errors; the read hangs
forever. `RUST_LOG=webrtc_sctp=debug` shows the same TSN in
`receive buffer full. dropping DATA with tsn=…` at every T3-rtx backoff. With
the `[patch.crates-io]` fork uncommented it delivers all 48 KiB. (The 32/48
KiB sizing just makes the trigger deterministic; the identical hole fires
probabilistically at the 1 MiB default.)

### Root cause and the principled fix

The receiver **advertises a horizon but enforces a byte count**. `a_rwnd` is
defined as a window beyond the cumulative TSN — a conforming sender keeps its
outstanding data inside it, and the dropped chunk here was within the window
the receiver itself advertised. Admission, however, is gated on "bytes
currently held in reassembly ≥ `max_receive_buffer_size`", a different
predicate; where the two disagree, the receiver refuses data it promised it
could hold — and can refuse exactly the chunk that would let it free
everything.

The principled fix is TCP-style window enforcement: always accept the
earliest outstanding data and shed the *latest* (highest-TSN) data when byte
pressure hits, so the head of line can always complete and the buffer bound
becomes the window itself. That is a larger change (admission, `a_rwnd`
computation and SACK generation all touch this accounting), so the patch
below is the minimal liveness version of the same invariant.

### Minimal fix (patch available)

When the payload queue is empty, accept the single in-sequence chunk
(`d.tsn == self.peer_last_tsn.wrapping_add(1)`): it is the only chunk that
can complete the pinned fragment sets and let the application drain. Because
it must carry the next cumulative TSN, this admits at most one chunk per
cumulative-ack advance, so the buffer bound is preserved against
non-conforming senders.

KPS uses this as a vendored patch:
https://github.com/privacy-ethereum/kps/commit/a73a0a8f9e4f76f3b89ba1155562ec836bcf3f3b

After the patch: 0/60 stalls in our stress loop vs ~2–3/60 before (0/100 in a
harsher run with an artificially small receive buffer). Happy to open a PR.

voltrevo:
Fixes https://github.com/webrtc-rs/webrtc/issues/822.

**Note**: This is the simple version of the fix. Instead of preventing the buffer from becoming full with unusable data, we allow it to overfill but only in the case that would otherwise deadlock.

PR per request, since the issue was still present in 0.20+.

> If not, could you submit a PR to master branch too?

### DIFF
--- a/rtc-sctp/src/association/mod.rs
+++ b/rtc-sctp/src/association/mod.rs
@@ -137,9 +137,9 @@ pub struct Association {
     bytes_received: usize,
     bytes_sent: usize,
 
-    peer_verification_tag: u32,
+    pub(crate) peer_verification_tag: u32,
     my_verification_tag: u32,
-    my_next_tsn: u32,
+    pub(crate) my_next_tsn: u32,
     peer_last_tsn: u32,
     // for RTT measurement
     min_tsn2measure_rtt: u32,
@@ -166,13 +166,13 @@ pub struct Association {
     local_addr: SocketAddr,
     transport_protocol: TransportProtocol,
 
-    source_port: u16,
-    destination_port: u16,
+    pub(crate) source_port: u16,
+    pub(crate) destination_port: u16,
     my_max_num_inbound_streams: u16,
     my_max_num_outbound_streams: u16,
     my_cookie: Option<ParamStateCookie>,
 
-    payload_queue: PayloadQueue,
+    pub(crate) payload_queue: PayloadQueue,
     inflight_queue: PayloadQueue,
     pending_queue: PendingQueue,
     control_queue: VecDeque<Packet>,
@@ -776,6 +776,10 @@ impl Association {
         self.max_message_size = max_message_size;
     }
 
+    fn has_readable_data(&self) -> bool {
+        self.streams.values().any(|s| s.reassembly_queue.is_readable())
+    }
+
     /// unregister_stream un-registers a stream from the association
     /// The caller should hold the association write lock.
     fn unregister_stream(&mut self, stream_identifier: StreamId, reason: AssociationError) {
@@ -1259,6 +1263,15 @@ impl Association {
                             self.payload_queue.push(d.clone(), self.peer_last_tsn);
                             stream_handle_data = true; //s.handle_data(d.clone());
                         }
+                    } else if d.tsn == self.peer_last_tsn.wrapping_add(1)
+                        && !self.has_readable_data()
+                    {
+                        debug!(
+                            "[{}] receive buffer full, but accepted as the in-sequence chunk (tsn={} ssn={}) that unblocks reassembly",
+                            self.side, d.tsn, d.stream_sequence_number
+                        );
+                        self.payload_queue.push(d.clone(), self.peer_last_tsn);
+                        stream_handle_data = true;
                     } else {
                         debug!(
                             "[{}] receive buffer full. dropping DATA with tsn={} ssn={}",
--- a/rtc-sctp/src/endpoint/endpoint_test.rs
+++ b/rtc-sctp/src/endpoint/endpoint_test.rs
@@ -2740,3 +2740,269 @@ fn test_association_shutdown_during_write() -> Result<()> {
 
     Ok(())
 }*/
+
+use crate::association::stream::StreamEvent;
+
+/// Deliver every queued inbound packet to the server ONE AT A TIME, and after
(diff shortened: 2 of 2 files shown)


rtc-sctp/src/association/mod.rs
```diff
@@ -1259,6 +1263,15 @@ impl Association {
                             self.payload_queue.push(d.clone(), self.peer_last_tsn);
                             stream_handle_data = true; //s.handle_data(d.clone());
                         }
+                    } else if d.tsn == self.peer_last_tsn.wrapping_add(1)
+                        && !self.has_readable_data()
+                    {
```
voltrevo, line 1268: The payload queue is empty and NOTHING is readable, so the buffer is full purely of incomplete fragment sets — and the only chunk that can ever complete one and let the application drain is this in-sequence chunk. Dropping it wedges the association at a permanent zero window (the sender retransmits it forever, every copy is dropped, and the application can neither see nor break the deadlock).

The `!has_readable_data()` guard is what keeps `max_receive_buffer_size` meaningful: a receiver whose application has merely stopped reading still has complete messages queued, so it is throttled exactly as before. Only a receiver that cannot hand the application anything accepts past the bound, and then by at most the remainder of one message.

rtc-sctp/src/association/mod.rs
```diff
@@ -137,9 +137,9 @@ pub struct Association {
     bytes_received: usize,
     bytes_sent: usize,
```
voltrevo, line 139: pub(crate) added for tests.

rainliu:
is this PR still under draft?

rainliu:
Thanks for raising this issue and submitted a PR to fix it.

I am going to close this PR in favor of another similar fix PR https://github.com/webrtc-rs/rtc/pull/218

PR #233, opened 2026-09-06
fix(sctp): measure maxPacketLifeTime from when a message is queued

### LINKED ISSUE #232: sctp: maxPacketLifeTime does not account for time a message spends in the pending queue
## Summary

`ReliabilityType::Timed` (the PR-SCTP policy that `maxPacketLifeTime` maps onto) measures a message's lifetime from the moment it is assigned a TSN, not from the moment the application wrote it. Time the message spends waiting in `pending_queue` — blocked on cwnd or a zero rwnd — is invisible to the policy.

The practical effect is that `maxPacketLifeTime` does not bound message age. After a congestion or connectivity stall clears, an association transmits messages that are arbitrarily older than their configured lifetime, and the lifetime clock only starts once they finally go out. For a live-media DataChannel this is the one case the option exists to prevent.

W3C describes `maxPacketLifeTime` as the window during which *transmissions and retransmissions* may occur. Today only the retransmission half is enforced.

## Environment

- `webrtc-rs/rtc`: `83c542e` (v0.21.0-rc.2)
- `webrtc-rs/webrtc`: `deaddd3` (v0.21.0-rc.2, `rtc` submodule at the above)

## Reproduction

Deterministic, on the existing `Pair` harness and its virtual clock — no sleeps. Add to `rtc-sctp/src/endpoint/endpoint_test.rs`:

```rust
#[test]
fn test_assoc_timed_reliability_ignores_time_spent_in_pending_queue() -> Result<()> {
    let si: u16 = 7;
    // maxPacketLifeTime = 100 ms.
    let lifetime_ms: u32 = 100;
    let stall = Duration::from_secs(1);

    let mut sbuf = vec![0u8; 1000];
    for i in 0..sbuf.len() {
        sbuf[i] = (i & 0xff) as u8;
    }

    let (mut pair, client_ch, server_ch) = create_association_pair(AckMode::NoDelay, 0)?;
    establish_session_pair(&mut pair, client_ch, server_ch, si)?;

    pair.client_stream(client_ch, si)?.set_reliability_params(
        true,
        ReliabilityType::Timed,
        lifetime_ms,
    )?;
    pair.server_stream(server_ch, si)?.set_reliability_params(
        true,
        ReliabilityType::Timed,
        lifetime_ms,
    )?;

    // Shrink the congestion window so exactly one message fits in flight. The
    // second write then has to wait in the pending queue behind the first.
    pair.client_conn_mut(client_ch).cwnd = 1200;

    // Message 0 leaves immediately and occupies the whole window.
    sbuf[0..4].copy_from_slice(&0u32.to_be_bytes());
    pair.client_stream(client_ch, si)?.write_sctp(
        &Bytes::from(sbuf.clone()),
        PayloadProtocolIdentifier::Binary,
    )?;

    // Message 1 is blocked behind it.
    sbuf[0..4].copy_from_slice(&1u32.to_be_bytes());
    pair.client_stream(client_ch, si)?.write_sctp(
        &Bytes::from(sbuf.clone()),
        PayloadProtocolIdentifier::Binary,
    )?;

    // Drive only the sender: message 0 goes on the wire, message 1 stays pending.
    // The peer is deliberately not driven, so nothing is acknowledged and the
    // window stays shut for as long as we like.
    pair.drive_client();
    assert!(
        pair.client_conn_mut(client_ch).buffered_amount() >= sbuf.len(),
        "message 1 should still be waiting in the pending queue"
    );

    // The stall lasts ten times the configured lifetime.
    pair.time += stall;

    // The stall clears and both sides run to completion.
    pair.drive();

    let mut buf = vec![0u8; 2000];
    let mut received = vec![];
    while let Some(chunks) = pair.server_stream(server_ch, si)?.read_sctp()? {
        chunks.read(&mut buf)?;
        received.push(u32::from_be_bytes([buf[0], buf[1], buf[2], buf[3]]));
    }

    assert!(
        !received.contains(&1),
        "message 1 waited {stall:?} in the pending queue with maxPacketLifeTime={lifetime_ms}ms \
         and must have been abandoned, but it was transmitted anyway (received={received:?})"
    );

    close_association_pair(&mut pair, client_ch, server_ch, si);

    Ok(())
}
```

The final assertion fires on current HEAD: message 1 is delivered despite having waited ten lifetimes in the pending queue.

## Evidence

The lifetime clock is reset when the chunk leaves the pending queue, in `move_pending_data_chunk_to_inflight_queue`:

```rust
// rtc-sctp/src/association/mod.rs:2914
c.since = Some(now); // use to calculate RTT and also for maxPacketLifeTime
```

`since` is dual-purpose — it is also the RTT baseline, which legitimately must be per-transmission (`rtc-sctp/src/chunk/chunk_payload_data.rs:116`). The Timed policy then measures against it:

```rust
// rtc-sctp/src/association/mod.rs:2803
} else if reliability_type == ReliabilityType::Timed {
    if let Some(since) = &c.since {
        let elapsed = now.duration_since(*since);
        if elapsed.as_millis() as u32 >= reliability_value {
            c.set_abandoned(true);
```

Nothing consults the lifetime before that point. `send_payload_data` pushes chunks into `pending_queue` with no timestamp (`mod.rs:2969`), and the drain loop in `pop_pending_data_chunks_to_send` checks only cwnd and rwnd (`mod.rs:2617`).

The path from the DataChannel option is intact, which is what makes this surprising in practice — `maxPacketLifeTime` is correctly negotiated and correctly reaches the policy (`rtc-datachannel/src/data_channel/mod.rs:328-329`); it just measures the wrong interval.

## Expected behavior

Per RFC 3758 timed reliability, a message whose lifetime has expired should be abandoned even if it has never been transmitted. A sender configured with `maxPacketLifeTime: N` should not put a message older than `N` ms on the wire.

## Proposed change

1. Record message creation time in a dedicated field on `ChunkPayloadData`, leaving `since` as the per-transmission RTT baseline.
2. Measure `ReliabilityType::Timed` against that field in `check_partial_reliability_status`.
3. Drop expired messages in `pop_pending_data_chunks_to_send` **before** cwnd/rwnd checks and before a TSN is assigned. Because no TSN has been allocated this needs no ForwardTSN and leaves no gap for the peer to reconcile.

Scope I would propose for a first change, to keep it provably safe:

- **Unordered streams only.** For an ordered stream the SSN is allocated in `packetize`, so dropping before transmission leaves an SSN gap that only a ForwardTSN carrying per-stream SSN can heal. Worth a follow-up, not this change.
- **Whole messages only**, all fragments together, and only at a message boundary — never a partial fragment run.
- **DCEP exempt**, matching the existing exemption in `check_partial_reliability_status`.
- **`reliability_value == 0` keeps current behaviour.** Expiring at the pending stage with a zero lifetime would mean never transmitting anything; I would rather not change what `maxPacketLifeTime: 0` means as a side effect of a bug fix. Happy to be told otherwise.

## Note on the sans-I/O time boundary

`send_payload_data` is reached from `Stream::write`, which has no `now`, so stamping creation time needs a time source. Per `docs/sans-io-deterministic-time.md` I would use a retained instant seeded in `Association::new` (which already takes `now`) and refreshed in each `handle_*` — the "last resort" row of that table, with no new ambient clock read and no allow-list change.

Worth noting that the caller's instant already exists one layer up and is simply dropped at the SCTP boundary. `RTCDataChannel::send` takes it:

```rust
// src/data_channel/mod.rs
pub fn send(&mut self, now: Instant, data: BytesMut) -> Result<()> {
    self.peer_connection.handle_write(TaggedRTCMessage { now, ... })
```

and it is still live in the handler:

```rust
// src/peer_connection/handler/datachannel.rs
fn handle_write(&mut self, msg: TaggedRTCMessageInternal) -> Result<()> {
    let now = msg.now;
    self.ctx.observe(now);
```

It is lost only because `sctp::Stream::write_sctp` does not accept one. So threading `now` through `Stream::write*` would not introduce a new time source — it would stop discarding an existing, already-sanctioned one (the `handle_write(msg)` row of that same table), and it would make the queueing timestamp exact rather than "as of the last time the association was told the time".

I did not do it that way because it is a breaking change to a public API and that is the maintainers' call. Happy to switch if preferred.

## Reference

- RFC 3758 — Partial Reliability Extension (timed reliability)
- RFC 8831 §6.6 — maps `maxPacketLifeTime` onto PR-SCTP timed reliability
- W3C WebRTC — `RTCDataChannelInit.maxPacketLifeTime`

realzoberg:
Fixes #232. Rebased onto `v0.21.x` on top of b1905e4, so this is now just the fix.

## What was wrong

`ReliabilityType::Timed` — the PR-SCTP policy `maxPacketLifeTime` maps onto — measured a message's lifetime from `c.since`, which is reassigned every time a chunk is transmitted:

```rust
// rtc-sctp/src/association/mod.rs
c.since = Some(now); // use to calculate RTT and also for maxPacketLifeTime
```

Time a message spends waiting in `pending_queue` for cwnd or rwnd was therefore invisible to the policy. Draining a stall transmitted messages arbitrarily older than their configured lifetime, each with a freshly reset clock, so only the retransmission half of the option ever worked.

## What this changes

b1905e4 already brings the caller's instant as far as `write_source`. This carries it the last hop into `packetize` and puts it to work:

- `ChunkPayloadData` gains `created_at`, stamped in `packetize` from that instant and never reassigned. `since` keeps its existing role as the per-transmission RTT baseline — the two really are different clocks.
- `check_partial_reliability_status` measures the `Timed` policy against `created_at`, using `saturating_duration_since`.
- `pop_pending_data_chunks_to_send` abandons expired messages ahead of the cwnd/rwnd checks and before a TSN is allocated. Because nothing has been assigned a TSN, this creates no gap for the peer to reconcile and needs no ForwardTSN. Buffered bytes are released through the existing `BufferedAmountReleased` / `BufferedAmountLow` path.

No public API change of its own, and no new clock reads:

```
$ python3 scripts/check-sans-io-deterministic-time.py
Sans-I/O deterministic-time boundary holds: 8 allow-listed reads, no new ones.
```

## Scope

Deliberately conservative for a first change:

- **Unordered messages only.** An ordered message has already consumed a stream sequence number in `packetize`, so abandoning it before it is ever transmitted would leave an SSN gap that only a ForwardTSN carrying that SSN can close. Worth a follow-up; I did not want to smuggle it in here.
- **Whole messages only**, every fragment, and only at a message boundary — a partially drained message is never truncated mid-run.
- **DCEP exempt**, matching the existing exemption in `check_partial_reliability_status`.
- **A zero lifetime keeps its current meaning** (transmit once, never retransmit) rather than silently becoming "never transmit at all". Happy to change this if you would rather `maxPacketLifeTime: 0` expire at the pending stage too.

## Tests

`test_assoc_timed_reliability_bounds_time_spent_in_pending_queue` is the regression test. It runs on the existing `Pair` harness and its virtual clock — no sleeps. Reverting the three source files and keeping the test reproduces the original failure on b1905e4:

```
message 1 waited 1s in the pending queue with maxPacketLifeTime=100ms and must have been
abandoned, but it was transmitted anyway (received=[0, 1])
```

With the fix:

```
$ cargo test -p rtc-sctp
test result: ok. 136 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo clippy
$ cargo fmt --all -- --check
```

`cargo test --workspace` is otherwise green. `ice_restart_by_webrtc_interop` and `mdns_query_and_gather_interop` fail on my machine both with and without this change — this host cannot bind IPv6 link-local addresses or run mDNS multicast — so they are environmental and unrelated.

### DIFF
--- a/rtc-sctp/src/association/mod.rs
+++ b/rtc-sctp/src/association/mod.rs
@@ -2597,6 +2597,81 @@ impl Association {
         self.bundle_data_chunks_into_packets(chunks, raw_packets);
     }
 
+    /// Whether the message at the head of the pending queue has outlived the
+    /// timed-reliability lifetime configured on its stream.
+    ///
+    /// Restricted to unordered messages: an ordered one has already consumed a
+    /// stream sequence number in `packetize`, so abandoning it before it is ever
+    /// transmitted would leave a gap that only a ForwardTSN carrying that SSN can
+    /// close. Restricted to a beginning fragment so a partially drained message is
+    /// never truncated mid-run.
+    fn pending_message_expired(&self, c: &ChunkPayloadData, now: Instant) -> bool {
+        if !c.unordered
+            || !c.beginning_fragment
+            || c.payload_type == PayloadProtocolIdentifier::Dcep
+        {
+            return false;
+        }
+
+        let Some(s) = self.streams.get(&c.stream_identifier) else {
+            return false;
+        };
+
+        // A zero lifetime keeps its established meaning â transmit once, never
+        // retransmit â rather than silently becoming "never transmit at all".
+        if s.reliability_type != ReliabilityType::Timed || s.reliability_value == 0 {
+            return false;
+        }
+
+        let Some(created_at) = c.created_at else {
+            return false;
+        };
+
+        now.saturating_duration_since(created_at).as_millis() as u32 >= s.reliability_value
+    }
+
+    /// Drop every fragment of the message at the head of the pending queue and
+    /// release the bytes it was holding in the stream's send buffer.
+    fn abandon_pending_message(&mut self, unordered: bool) {
+        let mut beginning_fragment = true;
+        let mut released = 0usize;
+        let mut stream_identifier = None;
+
+        loop {
+            let Some(c) = self.pending_queue.pop(beginning_fragment, unordered) else {
+                break;
+            };
+            beginning_fragment = false;
+            released += c.user_data.len();
+            stream_identifier = Some(c.stream_identifier);
+            if c.ending_fragment {
+                break;
+            }
+        }
+
+        let (Some(si), true) = (stream_identifier, released > 0) else {
+            return;
+        };
+
+        trace!(
+            "[{}] abandoned {} pending bytes on stream {} (maxPacketLifeTime)",
+            self.side, released, si
+        );
+
+        self.events
+            .push_back(Event::Stream(StreamEvent::BufferedAmountReleased {
+                id: si,
+                n_bytes: released,
+            }));
+
+        if let Some(s) = self.streams.get_mut(&si)
+            && s.on_buffer_released(released as i64)
+        {
+            self.events
+                .push_back(Event::Stream(StreamEvent::BufferedAmountLow { id: si }))
+        }
+    }
+
     /// pop_pending_data_chunks_to_send pops chunks from the pending queues as many as
     /// the cwnd and rwnd allows to send.
     fn pop_pending_data_chunks_to_send(
@@ -2631,6 +2706,16 @@ impl Association {
                     {
                         error!("[{}] failed to pop from pending queue", self.side);
                     }
+                    continue;
+                }
+
+                // RFC 3758 timed reliability: a message whose lifetime has run
+                // out must be abandoned even if it was never transmitted. Doing
+                // this here, ahead of the window checks and before a TSN is
+                // allocated, means there is no gap for the peer to reconcile and
+                // no ForwardTSN to send.
+                if self.pending_message_expired(c, now) {
+                    self.abandon_pending_message(unordered);
                     continue;
                 }
 
@@ -2801,8 +2886,11 @@ impl Association {
                     );
                 }
             } else if reliability_type == ReliabilityType::Timed {
-                if let Some(since) = &c.since {
-                    let elapsed = now.duration_since(*since);
+                // Age is measured from when the message was queued, not from its
+                // most recent transmission: `since` is the RTT baseline and is
+                // reassigned on every (re)transmission.
+                if let Some(since) = c.created_at.as_ref().or(c.since.as_ref()) {
+                    let elapsed = now.saturating_duration_since(*since);
                     if elapsed.as_millis() as u32 >= reliability_value {
                         c.set_abandoned(true);
                         trace!(
--- a/rtc-sctp/src/association/mod.rs
+++ b/rtc-sctp/src/association/mod.rs
@@ -2600,12 +2600,28 @@ impl Association {
     /// Whether the message at the head of the pending queue has outlived the
     /// timed-reliability lifetime configured on its stream.
     ///
+    /// Gated on `use_forward_tsn` for the same reason as
+    /// `check_partial_reliability_status`: when the peer did not negotiate
+    /// Forward-TSN the association falls back to reliable delivery, and partial
+    /// reliability must be off everywhere rather than in half the code.
+    ///
     /// Restricted to unordered messages: an ordered one has already consumed a
     /// stream sequence number in `packetize`, so abandoning it before it is ever
     /// transmitted would leave a gap that only a ForwardTSN carrying that SSN can
     /// close. Restricted to a beginning fragment so a partially drained message is
     /// never truncated mid-run.
+    ///
+    /// Only the head of the queue is inspected, so an expired message sitting
+    /// behind a fresh one keeps its buffered bytes until it reaches the head. It
+    /// is still never transmitted: it is re-checked once it gets there.
     fn pending_message_expired(&self, c: &ChunkPayloadData, now: Instant) -> bool {
+        if !self.use_forward_tsn {
+            return false;
+        }
+
+        // `packetize` forces `unordered = false` for DCEP, so the ordering test
+        // already excludes it; the explicit check keeps that non-obvious coupling
+        // from becoming load-bearing.
         if !c.unordered
             || !c.beginning_fragment
             || c.payload_type == PayloadProtocolIdentifier::Dcep
@@ -2632,21 +2648,38 @@ impl Association {
 
     /// Drop every fragment of the message at the head of the pending queue and
     /// release the bytes it was holding in the stream's send buffer.
-    fn abandon_pending_message(&mut self, unordered: bool) {
+    ///
+    /// Only called for an unordered message whose first fragment is at the head,
+    /// so the whole fragment run is contiguous: `send_payload_data` appends a
+    /// message's chunks without anything else interleaving. Running out before
+    /// the ending fragment would leave `PendingQueue` selected on an empty queue
+    /// and stall every later send, so that case asserts in debug builds and is
+    /// reported rather than silently tolerated.
+    fn abandon_pending_message(&mut self) {
         let mut beginning_fragment = true;
         let mut released = 0usize;
         let mut stream_identifier = None;
+        let mut completed = false;
 
-        loop {
-            let Some(c) = self.pending_queue.pop(beginning_fragment, unordered) else {
-                break;
-            };
+        while let Some(c) = self.pending_queue.pop(beginning_fragment, true) {
             beginning_fragment = false;
             released += c.user_data.len();
             stream_identifier = Some(c.stream_identifier);
             if c.ending_fragment {
+                completed = true;
                 break;
             }
+        }
+
+        if !completed {
+            debug_assert!(
+                false,
+                "pending queue ran out mid-message while abandoning an expired message"
+            );
+            error!(
+                "[{}] abandoned an incomplete fragment run; pending queue selection is now stale",
+                self.side
+            );
         }
 
         let (Some(si), true) = (stream_identifier, released > 0) else {
@@ -2715,7 +2748,7 @@ impl Association {
                 // allocated, means there is no gap for the peer to reconcile and
                 // no ForwardTSN to send.
                 if self.pending_message_expired(c, now) {
-                    self.abandon_pending_message(unordered);
+                    self.abandon_pending_message();
                     continue;
                 }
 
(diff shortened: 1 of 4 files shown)


realzoberg:
Follow-up on the "Time source" note in the description — I looked more closely at where the caller's instant actually goes, and the picture argues more strongly for the alternative than I gave it credit for.

The instant already exists one layer above SCTP and is simply discarded at the boundary. `RTCDataChannel::send` takes it from the application:

```rust
// src/data_channel/mod.rs
pub fn send(&mut self, now: Instant, data: BytesMut) -> Result<()> {
    self.peer_connection.handle_write(TaggedRTCMessage { now, ... })
```

and it is still live in the handler:

```rust
// src/peer_connection/handler/datachannel.rs
fn handle_write(&mut self, msg: TaggedRTCMessageInternal) -> Result<()> {
    let now = msg.now;
    self.ctx.observe(now);
```

It is lost only because `sctp::Stream::write_sctp` has no parameter for it.

So threading `now` through `Stream::write*` would not add a time source — it would stop discarding one that is already there and is already the sanctioned source for this path (`handle_write(msg)` → `msg.now`, per `docs/sans-io-deterministic-time.md`). The retained `last_now` this PR uses is the "last resort" row of that same table, and the doc is explicit about the cost:

> The retained instant is a last resort, not a shortcut. Inside a `handle_*` the caller's instant is right there — using a stored one instead reintroduces exactly the staleness this boundary exists to remove.

That applies here: with the retained instant a message is timestamped as of the last time the association happened to be told the time, not as of the `send` call. For a `maxPacketLifeTime` in the tens of milliseconds that error is bounded by the driver's polling cadence and should be small, but it is real, and it is avoidable.

I kept the retained instant because changing a public API signature is the maintainers' call, not mine. If you would rather have the exact version, say so and I will push it — it is a mechanical change to `write`, `write_sctp`, `write_with_ppi`, `write_chunk*` and their callers, and it would let `last_now` disappear entirely along with the `Option` it needed for `Default`.

rainliu (changes requested):
let's adopt "threading now through Stream::write*" to fix it. Please rebase PR to branch v0.21.x, which is ok for public API breaking change.

realzoberg:
Done — rebased onto `v0.21.x` and switched to threading `now` through `Stream::write*`. Thanks for the quick call on both.

- `write`, `write_sctp`, `write_with_ppi`, `write_chunk`, `write_chunk_with_ppi`, `write_chunks` now take `now: Instant`.
- `created_at` is stamped in `packetize` from that instant, so the timestamp is exact rather than "as of the last time the association was told the time".
- The retained `last_now` field is gone, and with it the `Option` it needed only because `Association` implements `Default`.
- Callers: `handler/sctp.rs` passes the `now` it already binds from `msg.now`; the `sctp_e2e` example passes its loop instant; tests use `pair.time` or their virtual-clock helper.

Gates on the final source: `cargo test -p rtc-sctp` 136 passed / 0 failed, `cargo clippy` clean, `cargo fmt --all -- --check` clean, and the boundary check still reports 8 allow-listed reads with no new ones.

`cargo test --workspace` is otherwise green. `ice_restart_by_webrtc_interop` and `mdns_query_and_gather_interop` fail on my machine with and without the change — this host cannot bind IPv6 link-local addresses or run mDNS multicast — so I have left them alone.

Two open questions from the description, both still yours to call: whether `maxPacketLifeTime: 0` should also expire at the pending stage, and whether to extend this to ordered streams (which needs a ForwardTSN carrying the SSN, so I kept it out of this change).

realzoberg:
Rebased onto `v0.21.x`.

Since b1905e4 carries the caller's instant as far as `write_source`, all this needed was the last hop into `packetize` plus the actual fix.

rtc-sctp/src/association/mod.rs
rainliu, line 3084: what about ChunkPayloadData with create_at: now?

rainliu: looks like _now: Instant, is redundant now since ChunkPayloadData is adding now in packetize, then we can remove this _now.

realzoberg:
Both addressed in 0694d95, thanks.

- `send_reset_request` now sets `created_at` on the EOS chunk from the instant you threaded in, so every locally created chunk carries one.
- `send_payload_data` no longer takes an instant at all — `packetize` stamps it, so the parameter was dead weight. Its only caller is updated.

While re-reading the change I also fixed four things:

- **Pending expiry is now gated on `use_forward_tsn`.** `check_partial_reliability_status` turns PR-SCTP off entirely when the peer did not negotiate Forward-TSN, and my pending-stage check did not inherit that, so a message could be dropped in a mode where the association is supposed to fall back to reliable delivery. 
- **`abandon_pending_message` could leave `PendingQueue` selected on an empty queue** if the fragment run ended early, which would stall every later send. Unreachable today because `send_payload_data` appends a message's chunks contiguously, so it now `debug_assert!`s and logs rather than silently corrupting the queue. The always-true `unordered` parameter is gone.
- The `c.since` comment still said it was the `maxPacketLifeTime` baseline, which directly contradicts the new field docs, and the `invalid c.since` error named the wrong field.
- Documented that only the head of the queue is inspected: an expired message queued behind a fresh one keeps its buffered bytes until it reaches the head. It is never transmitted — it is re-checked when it gets there — but `BufferedAmountLow` can lag in that configuration.

Tests are stronger too. The original assertion (`!received.contains(&1)`) would have passed a fix that dropped everything, so it now asserts the full delivered set and pins the buffered-byte release the change claims. And I added a fragmented-message case: at 1000 bytes the old test never entered the fragment loop in `abandon_pending_message`, which is the part that touches `PendingQueue`'s selected state. The new one abandons a 4000-byte message and then writes again to show the queue still works.

Both regression tests fail on b1905e4 and pass here:

```
left: [0, 1]           right: [0]
left: [1000, 4000, 700] right: [1000, 700]
```

One thing worth flagging that I got wrong in the description: "unordered messages only" describes the pending-queue drop, but moving the `Timed` baseline to `created_at` applies to every stream. For an ordered stream an expired message is now marked abandoned on its first transmission instead of after a full lifetime from that transmission — so it is still sent once, just no longer retransmitted. That is closer to RFC 3758 than before and not a regression, but it is a behaviour change and my wording implied ordered streams were untouched. Happy to add a CHANGELOG line, or to take the ordered case properly in a follow-up with the SSN in `fwd_tsn_stream_map`, whichever you prefer.

rainliu (approved):

PR #238, opened 2026-09-08
fix(sctp): enforce partial reliability across retransmission and reset

### LINKED ISSUE #237: sctp: Timed reliability does not actually stop retransmission at the lifetime.
With maxPacketLifeTime: 100, a message that is lost in flight is retransmitted by T3 and delivered to the peer two seconds later:

```rust
/// Under timed reliability the sender must stop transmitting a message once its
/// lifetime has expired. It does not: `abandoned` is only ever recomputed inside
/// `check_partial_reliability_status`, which never runs between the first
/// transmission and the T3 timeout, so the expired chunk still looks live when
/// `mark_all_to_retrasmit` inspects it.
#[test]
fn test_assoc_timed_reliability_does_not_retransmit_after_lifetime() -> Result<()> {
    let si: u16 = 9;
    let lifetime_ms: u32 = 100;
    let msg = Bytes::from_static(b"expires long before T3 can fire");

    let (mut pair, client_ch, server_ch) = create_association_pair(AckMode::NoDelay, 0)?;
    establish_session_pair(&mut pair, client_ch, server_ch, si)?;

    pair.client_stream(client_ch, si)?
        .set_reliability_params(true, ReliabilityType::Timed, lifetime_ms)?;
    pair.server_stream(server_ch, si)?
        .set_reliability_params(true, ReliabilityType::Timed, lifetime_ms)?;

    // Ignore whatever the handshake delivered.
    pair.server_conn_mut(server_ch).stats.reset();

    let now = pair.time;
    pair.client_stream(client_ch, si)?
        .write_sctp(now, &msg, PayloadProtocolIdentifier::Binary)?;

    // Transmit once, then lose it on the wire: the peer never sees it, and never
    // acknowledges it, so the retransmission timer is left to run.
    pair.drive_client();
    pair.server.inbound.clear();
    assert_eq!(
        0,
        pair.server_conn_mut(server_ch).stats.get_num_datas(),
        "the first transmission must have been lost for this test to mean anything"
    );

    // Well past the lifetime, still short of the one-second T3.
    pair.time += Duration::from_millis(500);
    pair.drive_client();

    // Let T3 fire, then deliver whatever the sender produced.
    pair.time += Duration::from_millis(1500);
    pair.drive_client();
    pair.drive_server();

    let redelivered = pair.server_conn_mut(server_ch).stats.get_num_datas();
    assert_eq!(
        0, redelivered,
        "a message whose maxPacketLifeTime is {lifetime_ms}ms expired at t+{lifetime_ms}ms and \
         must not be transmitted again, but the peer received it {redelivered} more time(s) \
         at t+2000ms"
    );

    Ok(())
}

assertion `left == right` failed: a message whose maxPacketLifeTime is 100ms expired
at t+100ms and must not be transmitted again, but the peer received it 1 more time(s)
at t+2000ms
  left: 0
 right: 1

test result: FAILED. 137 passed; 1 failed
```

The cause is that abandoned is only ever computed inside check_partial_reliability_status, which runs at three call sites — first transmission, the fast-retransmit path, and the T3 send path. None of them happens between the first transmission and the T3 timeout, so at T3:

1. mark_all_to_retrasmit (payload_queue.rs:153) tests abandoned(), sees false because the policy has never been evaluated for this chunk since it was sent, and marks the expired chunk for retransmission;
2. gather_data_packets_to_retransmit sends it;
3. check_partial_reliability_status runs only afterwards, at mod.rs:2574.

The policy is consulted after the decision instead of before it. RTO_MIN is 1000 ms, so this affects every lifetime shorter than a second.

The fast-retransmit path has the same ordering (mod.rs:2378 decides, mod.rs:2409 refreshes), but I have not reproduced that leg separately and don't want to claim it.

This looks like RFC 3758 territory rather than a new policy: under timed reliability the sender is not supposed to transmit a message after its lifetime has expired. The fix shape would be to evaluate the policy before committing to a retransmission rather than after.

realzoberg:
A lost DATA message with a 100 ms lifetime must not be retransmitted when T3 fires two seconds later. Check the queued message's immutable policy before assigning its first TSN and before every T3 or fast retransmission, including retries deferred by the send window. Closes #237.

A small deadline-only fix was not sufficient in the existing architecture. Reliability policy and buffer accounting depended on the live stream object, while stream reset was coupled to message delivery and SID reuse. Regression tests exposed lost policies after close, incorrect handling of late SACKs and fragmented messages, and incomplete reset recovery. Making #237 work through those transitions required separating message lifetime, peer acknowledgments, delivery ownership and RX/TX reset state. The PR therefore includes that internal restructuring and the tests for its dependencies; the public API is preserved.

Abandonment is one idempotent operation over a message identity: it covers sent fragments, advisory Gap-ACKed fragments and a pending tail, releases application buffer credit once, and retains the information needed to retransmit FORWARD-TSN. Actual peer acknowledgment, payload retention and congestion credit remain separate. Reliable messages and DCEP retain their reliability; `Timed(0)` preserves the existing API's initial-transmission behavior. These rules follow [RFC 3758 §§3.5 and 4.1](https://www.rfc-editor.org/rfc/rfc3758.html#section-3.5), [RFC 7496 §3.1](https://www.rfc-editor.org/rfc/rfc7496.html#section-3.1) and [RFC 9260 §§6–8](https://www.rfc-editor.org/rfc/rfc9260.html#section-6).

The close/reset dependencies are implemented explicitly. TX policy and SSN state outlive an API stream object. RX reset waits for its TSN boundary while complete unread messages remain in a separate delivery queue. An immutable outgoing reset stays serialized until its result is known: implicit receipt acknowledgment stops its timer without committing new SSNs, and recovery repeats the same request. Old DATA, EOF and the next DCEP OPEN pass through one lifecycle FIFO. See [RFC 6525 §5](https://www.rfc-editor.org/rfc/rfc6525.html#section-5) and [RFC 8831 §6.7](https://www.rfc-editor.org/rfc/rfc8831.html#section-6.7).

| Dependency and regression | Permanent coverage |
| --- | --- |
| Partial ACK → expiry → late ACK must release bytes once and keep T3 alive | `test_message_abandonment_is_idempotent_after_partial_and_late_acks`, `test_late_real_data_acks_keep_t3_alive`. |
| Lost fragments and FORWARD-TSN must not strand the pending tail | `test_timed_abandonment_covers_pending_tail_and_retries_forward_tsn`, `test_timed_zero_fragment_loss_restores_receive_window`. |
| `stop()` → write needs the reset SSN; another SID must keep sending | `test_write_after_stop_uses_reset_ssn`, `test_serialized_resets_and_reused_sid_do_not_block_other_streams`. |
| Implicit ACK and lost final result must not guess reset success | `test_same_reset_query_preserves_gates_timers_and_late_results`; external `forgotten_result_recovery` with both pinned peers. |
| Old messages, EOF and a new OPEN must reach the right DataChannel owner | `public_peer_reads_resume_saved_data_and_reset_without_network_input`, `old_dcep_ack_cannot_complete_new_fragmented_open`. |
| Duplicate/revoked Gap ACKs and TSN wrap must preserve recovery without scanning unrelated DATA | Eight `sack_recovery` scenarios; stale message selection after ACK and TSN reuse in `message_selection`. |
| A repeated Gap ACK must not repeat credit; a re-ACK after revocation must still update recovery | `duplicate_gap_ack_preserves_t3_and_does_not_repeat_credit`, `reack_after_gap_revocation_resets_t3_without_second_delivery_credit`. |
| Terminal abandonment must release retained payload storage before cumulative ACK | `final_abandonment_drops_payload_owner_before_cumulative_ack`; a separate tagged-allocation probe also checks physical backing release. |
| Gap-ACK overlap, cumulative advance, TSN wrap and close/reopen must not reuse stale state | `gap_range_overlap_matches_small_set_reference_across_cumulative_advance_and_wrap`, `closed_association_does_not_reuse_gap_ack_cache_after_cookie_echo`. |
| Due T3 must survive SACKs or new sends before timeout handling; actual earliest-DATA ACK still restarts it | Four `timer_deadline` regressions, including FORWARD-TSN-only recovery and repeated timeout at the same instant. |
| Releasing excess inflight capacity must preserve retained state, wraparound, revocation and byte credit | Two `shrink_test` regressions over busy, empty, physically wrapped and sparse queues. |

The PR is organized into six logical commits:

1. The SCTP/DataChannel fix, its regression tests and stateful tests.
2. Independent interop/performance tools and their CI workflow.
3. The targeted Gap-ACK and abandoned-payload ownership changes, with their regression tests.
4. Release read receive payloads and exhausted-Rexmit Gap-ACK backing while preserving recovery metadata, with ownership regression tests.
5. Avoid repeated Gap-ACK processing and SACK construction, and compact inflight storage, with independent boundary and queue-model tests.
6. Preserve overdue T3 recovery and release spare inflight storage after cumulative acknowledgments, with six regression tests.

### Performance

| Case | Master rate Mib/s | Updated PR rate Mib/s | Master RSS MiB | Updated PR RSS MiB | Master total CPU s/GiB | Updated PR total CPU s/GiB |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 KiB / N1 / ordered Reliable / default | 231.006 | 679.140 | 21.87 | 17.04 | 16.480 | 16.000 |
| 1 KiB / N1 / ordered Reliable / dedicated | 836.509 | 837.712 | 16.32 | 15.53 | 12.640 | 12.480 |
| 1 KiB / N1 / unordered Rexmit(0) / default | 1152.656 | 689.740 | 15.75 | 19.90 | 16.000 | 15.680 |
| 1 KiB / N1 / unordered Rexmit(0) / dedicated | 327.806 | 500.248 | 173.66 | 18.08 | 24.640 | 20.800 |
| 1 KiB / N10 / ordered Reliable / default | 2015.506 | 2249.952 | 97.54 | 70.04 | 22.848 | 21.056 |
| 1 KiB / N10 / ordered Reliable / dedicated | 755.495 | 803.647 | 159.50 | 128.86 | 12.784 | 11.904 |
| 1 KiB / N10 / unordered Rexmit(0) / default | 2336.300 | 2282.566 | 46.68 | 66.67 | 26.208 | 30.592 |
| 1 KiB / N10 / unordered Rexmit(0) / dedicated | 533.718 | 813.857 | 500.88 | 121.05 | 17.392 | 12.368 |
| 64 KiB / N1 / ordered Reliable / default | 151.821 | 165.147 | 23.93 | 23.12 | 8.480 | 7.840 |
| 64 KiB / N1 / ordered Reliable / dedicated | 287.806 | 287.319 | 14.07 | 12.78 | 8.640 | 8.640 |
| 64 KiB / N1 / unordered Rexmit(0) / default | 551.043 | 1070.900 | 174.04 | 25.18 | 18.560 | 8.000 |
| 64 KiB / N1 / unordered Rexmit(0) / dedicated | 517.075 | 1047.578 | 173.16 | 18.44 | 16.000 | 8.960 |
| 64 KiB / N10 / ordered Reliable / default | 2852.947 | 2855.815 | 106.45 | 103.73 | 15.792 | 14.976 |
| 64 KiB / N10 / ordered Reliable / dedicated | 980.695 | 989.540 | 105.77 | 77.85 | 8.416 | 8.064 |
| 64 KiB / N10 / unordered Rexmit(0) / default | 2537.431 | 2818.585 | 1041.89 | 121.05 | 26.384 | 21.088 |
| 64 KiB / N10 / unordered Rexmit(0) / dedicated | 851.305 | 1074.530 | 445.94 | 82.68 | 11.520 | 8.240 |

### DIFF
--- a/.github/workflows/sctp-interop.yml
+++ b/.github/workflows/sctp-interop.yml
@@ -1,0 +1,129 @@
+name: SCTP interoperability
+
+on:
+  push:
+    branches: [ master, v0.20.x, v0.21.x ]
+    paths:
+      - 'rtc-sctp/**'
+      - 'rtc-shared/**'
+      - 'rtc-datachannel/**'
+      - 'src/peer_connection/handler/mod.rs'
+      - 'src/peer_connection/handler/sctp.rs'
+      - 'src/peer_connection/handler/datachannel.rs'
+      - 'src/peer_connection/message/internal.rs'
+      - 'Cargo.toml'
+      - '.github/workflows/sctp-interop.yml'
+  pull_request:
+    branches: [ master, v0.20.x, v0.21.x ]
+    paths:
+      - 'rtc-sctp/**'
+      - 'rtc-shared/**'
+      - 'rtc-datachannel/**'
+      - 'src/peer_connection/handler/mod.rs'
+      - 'src/peer_connection/handler/sctp.rs'
+      - 'src/peer_connection/handler/datachannel.rs'
+      - 'src/peer_connection/message/internal.rs'
+      - 'Cargo.toml'
+      - '.github/workflows/sctp-interop.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  interop:
+    name: rtc with usrsctp and dcSCTP (both roles)
+    runs-on: ubuntu-24.04
+    timeout-minutes: 25
+    env:
+      SCTP_INTEROP_DIR: /tmp/sctp-interop
+      CARGO_TARGET_DIR: /tmp/sctp-interop-target
+      CARGO_TERM_COLOR: always
+      CC: clang
+      CXX: clang++
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@stable
+
+      - name: Prepare artifacts
+        run: mkdir -p "$SCTP_INTEROP_DIR/artifacts/logs" "$SCTP_INTEROP_DIR/artifacts/manifests"
+
+      - name: Install test-only native build tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes clang cmake
+
+      - name: Check Python scripts
+        run: |
+          python3 -m py_compile rtc-sctp/interop/*.py
+          python3 rtc-sctp/interop/test_build_usrsctp.py -v
+
+      - name: Build pinned usrsctp and run its adapter self-test
+        run: |
+          python3 rtc-sctp/interop/build_usrsctp.py --build-dir "$SCTP_INTEROP_DIR/usrsctp" --self-test \
+            2>&1 | tee "$SCTP_INTEROP_DIR/artifacts/logs/usrsctp-build.log"
+
+      - name: Build pinned dcSCTP and run its adapter self-test
+        run: |
+          python3 rtc-sctp/interop/build_dcsctp.py --build-dir "$SCTP_INTEROP_DIR/dcsctp" --self-test \
+            2>&1 | tee "$SCTP_INTEROP_DIR/artifacts/logs/dcsctp-build.log"
+
+      - name: Build the rtc public API adapter
+        run: |
+          cargo build -p rtc-sctp --example sctp_interop --release \
+            2>&1 | tee "$SCTP_INTEROP_DIR/artifacts/logs/rtc-build.log"
+
+      - name: Run all 40 packet scenarios
+        run: |
+          python3 rtc-sctp/interop/run.py \
+            --rtc "$CARGO_TARGET_DIR/release/examples/sctp_interop" \
+            --usrsctp "$SCTP_INTEROP_DIR/usrsctp/build/usrsctp_peer" \
+            --dcsctp "$SCTP_INTEROP_DIR/dcsctp/build/dcsctp-peer" \
+            --output "$SCTP_INTEROP_DIR/artifacts/traces" \
+            2>&1 | tee "$SCTP_INTEROP_DIR/artifacts/logs/scenarios.log"
+
+      - name: Collect build inputs and manifests
+        if: always()
+        run: |
+          mkdir -p "$SCTP_INTEROP_DIR/artifacts/manifests"
+          git rev-parse HEAD > "$SCTP_INTEROP_DIR/artifacts/manifests/rtc-commit.txt"
+          rustc --version --verbose > "$SCTP_INTEROP_DIR/artifacts/manifests/rust-toolchain.txt" 2>&1 || true
+          python3 --version > "$SCTP_INTEROP_DIR/artifacts/manifests/python-version.txt" 2>&1 || true
+          clang --version > "$SCTP_INTEROP_DIR/artifacts/manifests/compiler.txt" 2>&1 || true
+          cmake --version > "$SCTP_INTEROP_DIR/artifacts/manifests/cmake-version.txt" 2>&1 || true
+          python3 - <<'PY'
+          import os
+          from pathlib import Path
+          import shutil
+
+          root = Path(os.environ["SCTP_INTEROP_DIR"])
+          output = root / "artifacts/manifests"
+          files = {
+              Path("Cargo.lock"): "Cargo.lock",
+              root / "usrsctp/build/build.json": "usrsctp-build.json",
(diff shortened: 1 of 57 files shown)


rainliu (changes requested):
This PR is not reviewable.

I suggest split the PR into 5 PRs:

1. The fix to issue #237  — policy refresh before C2/mark_all_to_retrasmit at T3, the same ordering correction in the fast-retransmit leg, plus regression test and a fast-retransmit counterpart.
2. Rexmit(0) timing — the RFC 7496 §3.1 correction with its own tests and a note on the memory tradeoff. 
3. Reset / SID-reuse lifecycle — the reset*.rs and handler/* work.
4. Perf — already separate commits; they rebase onto 1–3 cleanly as their own PR with the benchmark numbers attached.
5.  usrsctp/dcSCTP Interop harness + CI

PR #239, opened 2026-09-12
feat(dtls): allow a server config to hold both a PSK callback and certificates

nmammeri:
## Problem

A DTLS server config could hold a PSK callback *or* certificates, never both: `validate` rejected the combination outright, and the flights decided which credential path a handshake took by asking the config (`cfg.local_psk_callback.is_none()`).

That forces one listening socket per credential type. A deployment whose peers are provisioned with a mixture of PSKs and certificates needs a separate server for each.

## Change

A server config may now carry both, and which credential a given handshake uses is settled by the cipher suite it negotiated rather than by what the config happens to hold.

- **`validate`**: `ErrPskAndCertificate` now applies to clients only. A client presents one credential, so for it the two remain exclusive; a server lets each handshake pick.
- **`State::is_cipher_suite_psk()`**: new accessor for the negotiated suite's family. flight0 and flight4 use it in place of `cfg.local_psk_callback.is_none()` when choosing the named curve, the ServerHello extensions, and whether to send a Certificate. The suite is chosen from the ClientHello in flight 0, so it is known at every point below.
- **`ConfigBuilder::offered_cipher_suites`**: centralises which families go into the offer. A client drops the certificate suites when it holds a PSK; a server drops them only when it holds no certificate — having a PSK callback is no longer the answer to whether it has a certificate.

Client behaviour is unchanged: it offers one family or the other, and its own certificate is optional either way.

## Test

`one_server_config_serves_psk_and_certificate_clients` builds a single server config holding both a PSK callback and a certificate, then drives a full handshake and record exchange against a PSK client and a certificate client in turn.

## Verification

`cargo test --workspace --all-features` — 1898 passed, 0 failed.
`cargo clippy --workspace --all-features --all-targets` — no new warnings.

### DIFF
--- a/rtc-dtls/src/config.rs
+++ b/rtc-dtls/src/config.rs
@@ -367,6 +367,18 @@ pub enum ExtendedMasterSecretType {
 }
 
 impl ConfigBuilder {
+    fn offered_cipher_suites(&self, is_client: bool) -> Result<Vec<Box<dyn CipherSuite>>> {
+        // a client offers one family or the other, while a server can only offer them if
+        // it actually holds a certificate.
+        let exclude_certificate = if is_client {
+            self.psk.is_some()
+        } else {
+            self.certificates.is_empty()
+        };
+
+        parse_cipher_suites(&self.cipher_suites, self.psk.is_none(), exclude_certificate)
+    }
+
     fn validate(&self, is_client: bool) -> Result<()> {
         if is_client && self.psk.is_some() && self.psk_identity_hint.is_none() {
             return Err(Error::ErrPskAndIdentityMustBeSetForClient);
@@ -376,7 +388,8 @@ impl ConfigBuilder {
             return Err(Error::ErrServerMustHaveCertificate);
         }
 
-        if !self.certificates.is_empty() && self.psk.is_some() {
+        // A server may hold both and let each handshake pick
+        if is_client && !self.certificates.is_empty() && self.psk.is_some() {
             return Err(Error::ErrPskAndCertificate);
         }
 
@@ -384,7 +397,7 @@ impl ConfigBuilder {
             return Err(Error::ErrIdentityNoPsk);
         }
 
-        parse_cipher_suites(&self.cipher_suites, self.psk.is_none(), self.psk.is_some())?;
+        self.offered_cipher_suites(is_client)?;
 
         Ok(())
     }
@@ -405,12 +418,12 @@ impl ConfigBuilder {
         })?;
         self.validate(is_client)?;
 
-        let mut local_cipher_suites: Vec<CipherSuiteId> =
-            parse_cipher_suites(&self.cipher_suites, self.psk.is_none(), self.psk.is_some())?
-                .iter()
-                .map(|cs| cs.id())
-                .filter(|id| id.supported_by(crypto_provider.crypto()))
-                .collect();
+        let mut local_cipher_suites: Vec<CipherSuiteId> = self
+            .offered_cipher_suites(is_client)?
+            .iter()
+            .map(|cs| cs.id())
+            .filter(|id| id.supported_by(crypto_provider.crypto()))
+            .collect();
         if local_cipher_suites.is_empty() {
             return Err(Error::ErrNoAvailableCipherSuites);
         }
--- a/rtc-dtls/src/endpoint.rs
+++ b/rtc-dtls/src/endpoint.rs
@@ -336,27 +336,32 @@ mod tests {
         }
     }
 
+    /// Builds a config offering `suites`, with whichever credentials they need — pass both
+    /// families to get a server holding a psk callback and certificates both.
     fn config(
         provider: Arc<dyn RTCCryptoProvider>,
         is_client: bool,
-        suite: CipherSuiteId,
+        suites: &[CipherSuiteId],
     ) -> Result<Arc<HandshakeConfig>> {
         let mut builder = ConfigBuilder::default()
             .with_crypto_provider(provider.clone())
-            .with_cipher_suites(vec![suite])
+            .with_cipher_suites(suites.to_vec())
             .with_insecure_skip_verify(true);
-        let is_psk = matches!(
-            suite,
-            CipherSuiteId::Tls_Psk_With_Aes_128_Ccm
-                | CipherSuiteId::Tls_Psk_With_Aes_128_Ccm_8
-                | CipherSuiteId::Tls_Psk_With_Aes_128_Gcm_Sha256
-        );
-        if is_psk {
+        let is_psk = |suite: &CipherSuiteId| {
+            matches!(
+                suite,
+                CipherSuiteId::Tls_Psk_With_Aes_128_Ccm
+                    | CipherSuiteId::Tls_Psk_With_Aes_128_Ccm_8
+                    | CipherSuiteId::Tls_Psk_With_Aes_128_Gcm_Sha256
+            )
+        };
+        if suites.iter().any(is_psk) {
             builder = builder.with_psk(Some(Arc::new(|_| Ok(vec![0xab, 0xcd, 0xef]))));
             if is_client {
                 builder = builder.with_psk_identity_hint(Some(b"rtc-dtls-test".to_vec()));
             }
-        } else if !is_client {
+        }
+        if !is_client && suites.iter().any(|suite| !is_psk(suite)) {
             builder = builder.with_certificates(vec![Certificate::generate_self_signed(
                 vec!["localhost".to_owned()],
                 provider.crypto(),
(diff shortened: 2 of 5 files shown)


rtc-dtls/src/flight/flight0.rs
```diff
@@ -196,7 +196,7 @@ impl Flight for Flight0 {
         state.remote_epoch = 0;
 
         state.named_curve = DEFAULT_NAMED_CURVE;
-        if cfg.local_psk_callback.is_none() {
+        if !state.is_cipher_suite_psk() {
             state.named_curve = cfg.local_named_curves[0];
```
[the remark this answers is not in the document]
rainliu: The PR description says "the suite is chosen from the ClientHello in flight 0, so it is known at every point below." True for flight 4, false for flight 0's own generate. The FSM is Preparing → generate, then Waiting → parse (handshaker.rs:96-122, 182), so state.cipher_suite is still None here. I probed it:
```
PROBE flight0::generate cipher_suite_set=false is_psk=false
PROBE flight0::generate cipher_suite_set=false is_psk=false
test endpoint::tests::one_server_config_serves_psk_and_certificate_clients ... ok
```
Both handshakes — including the PSK one. So !is_cipher_suite_psk() is unconditionally true, and this replaced a check that was meaningful. Two consequences:

- A PSK-only server's initial named_curve silently flips from DEFAULT_NAMED_CURVE (X25519) to local_named_curves[0] (P256).
- cfg.local_named_curves[0] now runs for PSK configs. config.rs:443 only rejects an empty curve list when psk.is_none(), so a PSK config on a provider advertising no key-exchange support panics on index 0 where it previously didn't. Only reachable via a custom RTCCryptoProvider, but that's the pluggable path the crate exists to support.

Since the suite genuinely isn't known at this point, the honest fix is to stop pretending otherwise — keep the config-based test here, or just make the index safe:
```
state.named_curve = DEFAULT_NAMED_CURVE;
if let Some(curve) = cfg.local_named_curves.first() {
    state.named_curve = *curve;
}
```
Flight0::parse overwrites this from the ClientHello's supported-curves anyway, so the line only matters when the client offers nothing supported.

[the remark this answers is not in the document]
nmammeri: Good catch, code reverted.

rtc-dtls/src/config.rs
```diff
@@ -367,6 +367,18 @@ pub enum ExtendedMasterSecretType {
...
+    fn offered_cipher_suites(&self, is_client: bool) -> Result<Vec<Box<dyn CipherSuite>>> {
+        // a client offers one family or the other, while a server can only offer them if
+        // it actually holds a certificate.
+        let exclude_certificate = if is_client {
+            self.psk.is_some()
+        } else {
+            self.certificates.is_empty()
+        };
+
+        parse_cipher_suites(&self.cipher_suites, self.psk.is_none(), exclude_certificate)
```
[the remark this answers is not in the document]
nmammeri: The signature, curve and signing-key checks in `build` fixed to cater for mixed config properly by gating certificate checks on the offer, not the psk callback for the server side.

rtc-dtls/src/endpoint.rs
```diff
@@ -486,12 +494,33 @@ mod tests {
...
+            CipherSuiteId::Tls_Ecdhe_Ecdsa_With_Aes_128_Gcm_Sha256,
+            CipherSuiteId::Tls_Psk_With_Aes_128_Ccm_8,
+        ];
+
+        for client_suite in certificate_and_psk_suites {
+            handshake_and_exchange(
+                provider.clone(),
+                provider.clone(),
+                client_suite,
+                &certificate_and_psk_suites,
```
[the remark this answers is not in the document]
nmammeri: Test changed as suggested.

rtc-dtls/src/config.rs
rainliu: the certificate/signature-family filter is now skipped for dual configs
```
if !is_client && self.psk.is_none() {
    let signing_key = &self.certificates[0].private_key.signing_key;
    local_cipher_suites.retain(/* drop ECDHE suites this key can't sign */);
```
psk.is_none() was a correct proxy for "has certificates" only while the two were mutually exclusive — which is exactly what this PR removes. A server holding both now skips the retain, so a dual config with an RSA certificate still advertises TLS_ECDHE_ECDSA_*, and the mismatch doesn't surface until select_signature_scheme / generate_key_signature in flight 4 builds a ServerKeyExchange whose signature family contradicts the negotiated suite. Same guard, same problem at :429 (ErrNoAvailableSignatureSchemes) and :443 (ErrNoAvailableCipherSuites).

All three should key off !self.certificates.is_empty() instead. The new test uses an ECDSA self-signed cert, so it can't catch this — this is the one place where "server may hold both" genuinely needs new validation logic rather than a relaxed check.

rtc-dtls/src/flight/flight4.rs
```diff
@@ -290,7 +290,9 @@ impl Flight for Flight4 {
                 }
 
                 let mut pre_master_secret = vec![];
-                if let Some(local_psk_callback) = &cfg.local_psk_callback {
+                if state.is_cipher_suite_psk()
```
rainliu: silent-fallthrough
```
if state.is_cipher_suite_psk() && let Some(local_psk_callback) = &cfg.local_psk_callback {
```
PSK suite + absent callback now falls into the else if let Some(local_keypair) arm and derives the pre-master secret from the client's identity hint as though it were an EC point — or, with no keypair either, leaves pre_master_secret as vec![] and derives a master secret from nothing, with no error. Not reachable today (a config without a PSK never offers PSK suites), but an explicit else returning ErrPskAndIdentityMustBeSetForClient-style failure costs less than the reasoning needed to prove it stays unreachable.

rainliu (changes requested):
test_psk_configuration in conn/conn_test.rs:1553-1557 asserts ErrPskAndCertificate for both client and server, which this PR makes wrong for the server. It doesn't fail because lines 50–2365 of that file are inside a block comment. Worth updating anyway so it isn't re-enabled onto a stale expectation — and nothing live asserts the client-side rejection the PR deliberately keeps, which is a cheap validate unit test.

rtc-dtls/src/flight/flight4.rs
nmammeri: Fixed with an explicit error

rtc-dtls/src/config.rs
nmammeri: already addressed

rtc-dtls/src/config.rs
```diff
@@ -367,6 +367,18 @@ pub enum ExtendedMasterSecretType {
 }
 
 impl ConfigBuilder {
+    fn offered_cipher_suites(&self, is_client: bool) -> Result<Vec<Box<dyn CipherSuite>>> {
+        // a client offers one family or the other, while a server can only offer them if
+        // it actually holds a certificate.
+        let exclude_certificate = if is_client {
+            self.psk.is_some()
+        } else {
+            self.certificates.is_empty()
```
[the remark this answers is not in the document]
nmammeri: docs updated

rtc-dtls/src/cipher_suite/mod.rs
```diff
@@ -180,6 +180,15 @@ impl From<&str> for CipherSuiteId {
 }
 
 impl CipherSuiteId {
+    pub(crate) fn is_psk(self) -> bool {
```
rainliu: There are now two independent definitions of "is this a PSK suite": the existing CipherSuite::is_psk() trait method, implemented per suite, and the new inherent CipherSuiteId::is_psk() matching on three variants. Adding a fourth PSK suite means remembering both.

Suggest not adding it, instead, directly using CipherSuite::is_psk()

rtc-dtls/src/config.rs
```diff
@@ -405,12 +419,12 @@ impl ConfigBuilder {
         })?;
         self.validate(is_client)?;
 
-        let mut local_cipher_suites: Vec<CipherSuiteId> =
-            parse_cipher_suites(&self.cipher_suites, self.psk.is_none(), self.psk.is_some())?
-                .iter()
-                .map(|cs| cs.id())
-                .filter(|id| id.supported_by(crypto_provider.crypto()))
-                .collect();
+        let mut local_cipher_suites: Vec<CipherSuiteId> = self
```
rainliu: you can change it to Vec<(CipherSuiteId, is_psk)>

rainliu:
In addition, add test coverage for the preserved client-side rejection. Still no live test asserting ErrPskAndCertificate for a client, and conn_test.rs:1556-1557 still expects it for both sides

rtc-dtls/src/config.rs
nmammeri: refactored as suggested

rtc-dtls/src/cipher_suite/mod.rs
nmammeri: removed

nmammeri:
test asserting ErrPskAndCertificate is added but  conn_test.rs:1556-1557 is commented out.

rainliu (approved):

PR #242, opened 2026-09-15
fix(sctp): defer Rexmit abandonment until retransmission

realzoberg:
`Rexmit(0)` currently marks DATA abandoned on its first transmission, so a later SACK can trigger FORWARD-TSN before any retransmission is due. This PR keeps the message outstanding until it is acknowledged or would need a retry, as required by [RFC 7496 §3.1](https://www.rfc-editor.org/rfc/rfc7496.html#section-3.1). Initial transmission does not consume the retry allowance; fast and T3 retries share that allowance.

The queued message owns its immutable policy. T3 refreshes policy before C2 and marking retries. Deferred T3 and fast recovery then check only the candidates being considered for retransmission, including their current deadline. Exhaustion retires the complete message: retained fragments, Gap-ACKed siblings and any pending tail. Earlier fragments selected into the same batch are removed before attempts are committed. The fragment walk handles cumulative-ACKed prefixes and TSN wrap; payload credit is released once and peer receipt stays distinct from abandonment, following [RFC 3758 §3.5](https://www.rfc-editor.org/rfc/rfc3758.html#section-3.5).

T3 retains congestion-control actions and DATA RTO backoff even when all candidates are abandoned. T2-shutdown and RECONFIG take the current DATA RTO when started, then double their own interval once per expiry, capped at RTO.Max. Rearming after packet construction does not compound either control timer's backoff with T3. RECONFIG deliberately keeps the current DATA RTO as its initial value, as specified by [RFC 6525 §5.1.1](https://www.rfc-editor.org/rfc/rfc6525.html#section-5.1.1). `Association::rtt()` reports SRTT independently of timeout backoff, with documented zero before the first measurement.

Pending-message removal validates the fragment boundary and accounting before dequeuing. A stale position remains an idempotent no-op. Malformed queue state returns an error and closes the association through the existing transport-error path, without emitting a partially selected batch or rearming timers.

Memory tradeoff: unacknowledged `Rexmit(0)` payload remains buffered while waiting for ACK or a recovery event, increasing retention relative to eager abandonment. Abandonment returns application buffer credit once; TSN/SSN metadata remains until cumulative acknowledgment, and retained records can still hold physical backing storage.

Recovery measurements compare the previous PR head `d0fdd5c` with this change using the identical `benchmark_recovery_flushes` test and Cargo.lock, Rust 1.92.0 on macOS arm64, seven alternating pairs after warmup. Each run recovers the whole window one DATA chunk per flush. Setup and the initial T3 marking pass are excluded; cumulative progress is applied directly to isolate recovery selection/serialization from SACK parsing and congestion control. Queue lookups are counted in separate instrumented runs. These are recovery-path measurements, not end-to-end network throughput.

| Recovery | Window (chunks) | Median before → after (ms) | Queue lookups before → after |
| --- | ---: | ---: | ---: |
| T3 | 128 | 0.105 → 0.079 | 8,640 → 1,022 |
| Fast | 128 | 0.107 → 0.077 | 8,896 → 1,278 |
| T3 | 1,024 | 2.529 → 0.599 | 527,872 → 8,190 |
| Fast | 1,024 | 2.623 → 0.600 | 529,920 → 10,238 |
| T3 | 4,096 | 40.273 → 2.486 | 8,402,944 → 32,766 |
| Fast | 4,096 | 41.416 → 2.551 | 8,411,136 → 40,958 |

Relative MAD is 1.4–4.3% except the updated 4,096-chunk fast case at 5.25%; that timing is noisier. Lookup counts are identical across all seven samples. Run the measurement with `cargo test -p rtc-sctp --lib --release benchmark_recovery_flushes -- --ignored --nocapture`. The ordinary regression suite also enforces a fixed lookup bound for one retry across increasing window sizes.

Validation: 173 SCTP tests pass in debug and release; 279 RTC and 30 DataChannel library tests pass, along with the DataChannel backpressure and SCTP receive-window integration tests. The manual benchmark is ignored in ordinary test runs and was executed separately. Regression checks for the full-window scan, T2 compounding, RTT reporting and malformed pending messages fail on `d0fdd5c` and pass with the fixes. Formatting, the deterministic-time check and `cargo clippy -p rtc-sctp --lib -- -D warnings` pass. Strict Clippy across all targets still reports the same 58 pre-existing test diagnostics, with no added diagnostics.


rtc-sctp/src/association/mod.rs
```diff
@@ -2564,6 +2511,9 @@ impl Association {
     /// get_data_packets_to_retransmit is called when T3-rtx is timed out and retransmit outstanding data chunks
     /// that are not acked or abandoned yet.
     fn get_data_packets_to_retransmit(&mut self, now: Instant, raw_packets: &mut Vec<Bytes>) {
+        // T3 may have marked this DATA before its lifetime elapsed, while the
+        // send window deferred the actual retransmission (RFC 3758 TR4).
+        self.abandon_unretransmittable_messages(now, |c| c.retransmit);
```
rainliu, line 2516: O(W²) rescan during T3 recovery:

get_data_packets_to_retransmit now opens with abandon_unretransmittable_messages(now, |c| c.retransmit), and unretransmittable_messages walks every TSN in inflight_queue.sorted with a chunk_map probe each. It runs on every gather_outbound while t3_retransmit_pending is set. With a 1 MB window (~870 chunks), T3 collapses cwnd to one MTU; each gather then retransmits ~1 chunk, sets chunks_remaining, and keeps the flag true — so the full window is rescanned once per outbound flush for the whole recovery. The code it replaces (refresh_partial_reliability_status) was explicitly documented "Run on T3 only … the timer fires at most once per RTO", and the adjacent create_forward_tsn comment records that an identical O(rwnd) rescan "dominated the send profile." Hoist the scan to the T3 handler, or bound it to the chunks actually under consideration.

rtc-sctp/src/association/mod.rs
```diff
@@ -2368,6 +2346,10 @@ impl Association {
     ) -> Vec<Bytes> {
         if self.will_retransmit_fast {
             self.will_retransmit_fast = false;
+            self.abandon_unretransmittable_messages(
```
rainliu, line 2349: Same scan per SACK in fast recovery:

will_retransmit_fast is set by process_fast_retransmission on every SACK that advances the cumulative ACK point while in_fast_recovery, so gather_outbound_fast_retransmission_packets does one full sorted + chunk_map traversal per SACK for the duration of recovery. Pre-PR, the reliability policy was evaluated only for chunks that had already cleared the cheap nsent > 1 || miss_indicator < 3 disqualifiers — so this is new per-SACK cost proportional to the congestion window.

rtc-sctp/src/association/mod.rs
```diff
@@ -3267,6 +3218,9 @@ impl Association {
 
             Timer::T3RTX => {
                 self.stats.inc_t3timeouts();
+                self.rto_mgr.backoff();
```
rainliu, line 3221: Compounding backoff for T2-shutdown:

The new rto_mgr.backoff() doubles the shared RTO on every T3 expiry, and TimerTable::start exempts only Ack and T3RTX from calculate_next_timeout(interval, retrans[timer]). T2-shutdown is started from rto_mgr.get_rto() and is still subject to that. In ShutdownSent the association still runs gather_data_packets_to_retransmit, so T3 can expire; after three backoffs the next T2-shutdown arm is 8·RTO << retrans[T2Shutdown] — both backoffs compounding, well past RFC 9260's 2ⁿ·RTO. This is the same bug class the PR deliberately fixed for Timer::Reconfig via reconfig_interval; T2-shutdown was left behind (as is the Reconfig seed, taken from an already-backed-off RTO). Separately, the public Association::rtt() now reports the backed-off RTO rather than an RTT estimate — unused in-tree, but it is public API.

rtc-sctp/src/queue/pending_queue.rs
```diff
@@ -49,6 +58,39 @@ impl PendingQueue {
...
+    /// the next message, even when it has the same stream and policy metadata.
+    pub(crate) fn drain_message(&mut self, position: PendingPosition) -> Vec<ChunkPayloadData> {
+        if self.front_position() != Some(position) {
+            return vec![];
+        }
+        let mut chunks = vec![];
+        loop {
+            let c = self
+                .peek()
+                .expect("pending message must have an ending fragment");
```
rainliu, line 83: never Panic!

Avoid .expect() and .unwrap().

explicitly handle them.

rainliu (changes requested):

rainliu (approved):

PR #249, opened 2026-09-21
refactor(rtc): drop SctpTransport::internal_buffer

threeseed:
`internal_buffer` was only ever read for its length, which is the negotiated max message size. Use that directly and drop a zero-filled allocation of up to 256 KiB per connection. No behaviour change.


threeseed:
Folded into #251.

PR #250, opened 2026-09-21
fix(rtc): bound inbound SCTP messages by the local max message size

threeseed:
Builds on #249; only the last commit is new.

Inbound reassembly was bounded by the negotiated send limit, the smaller of the two peers' limits. Per [RFC 8841 §6.1](https://datatracker.ietf.org/doc/html/rfc8841#section-6.1) the peer may send anything up to the `max-message-size` this endpoint advertises, so a peer that advertised less than it sends (or omitted the attribute, implying 64 KiB) had larger messages dropped with `ErrShortBuffer`. Inbound messages are now bounded by the local limit.


threeseed:
Folded into #251.

